PR #21, opened 2021-09-05
support for eip1559

alexshchur:


src/transaction-builder.ts
```diff
@@ -40,7 +42,9 @@ export class TransactionBuilder {
     userAddress: Address;
     partnerAddress: Address;
     partnerFeePercent: string;
-    gasPrice: string;
+    gasPrice?: string; // // @TODO: improve types? so that either gasPrice or ALL of max.*FeePerGas MUST be returned?
```
mwamedacen, line 45: I think this is important enough to handle in this iteration. Basically extracting this inlined type into its own interface and having sub interfaces legacy / eip1559 should do the trick :)

mwamedacen: But I'm thinking... this won't be statically checked anyway as it'll be passed through from requests from api to here so  at runtime. The best solution might be handling it in validator schema.

mwamedacen: And you did it https://github.com/paraswap/paraswap-api/blob/1840bb8f026514053d4a31dc55f3719d71125b1a/src/lib/validators.ts#L198 !

mwamedacen (approved):

shresthagrawal (approved):

PR #25, opened 2021-09-20
feat: kyber dmm

nduchak:
Closes BACK-99

### DIFF
--- a/src/dex/kyberdmm.ts
+++ b/src/dex/kyberdmm.ts
@@ -27,7 +27,7 @@ type KyberDmmParam = [
 
 enum KyberDMMFunctions {
   swapExactTokensForTokens = 'swapExactTokensForTokens',
-  swapExactTokensForETH = 'swapExactTokensForETH',
+  swapTokensForExactTokens = 'swapTokensForExactTokens',
 }
 
 export class KyberDmm
@@ -62,7 +62,7 @@ export class KyberDmm
           path: 'address[]',
         },
       },
-      { poolPath: [data.pools[0].address], path: data.path },
+      { poolPath: data.pools.map(p => p.address), path: data.path },
     );
     return {
       targetExchange: data.router,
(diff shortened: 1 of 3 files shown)


src/dex/kyberdmm.ts
```diff
@@ -0,0 +1,104 @@
...
+    destToken: string,
+    srcAmount: string,
+    destAmount: string,
+    data: KyberDmmData,
+    side: SwapSide,
+  ): Promise<SimpleExchangeParam> {
+    const swapFunctionParams: KyberDmmParam = [
+      srcAmount,
+      destAmount,
+      [data.pools[0].address],
```
shresthagrawal, line 85: Shouldn't this be?
```suggestion
      data.pools.map(p => p.address),
```

src/dex/kyberdmm.ts
```diff
@@ -0,0 +1,104 @@
...
+    side: SwapSide,
+  ): AdapterExchangeParam {
+    const payload = this.abiCoder.encodeParameter(
+      {
+        ParentStruct: {
+          poolPath: 'address[]',
+          path: 'address[]',
+        },
+      },
+      { poolPath: [data.pools[0].address], path: data.path },
```
shresthagrawal, line 65: similar to the comment below
```suggestion
      { poolPath: data.pools.map(p => p.address), path: data.path },
```

src/dex/kyberdmm.ts
```diff
@@ -0,0 +1,104 @@
...
+  destToken: string,
+  pools: string[],
+  path: string[],
+  to: string,
+  deadline: string,
+];
+
+enum KyberDMMFunctions {
+  swapExactTokensForTokens = 'swapExactTokensForTokens',
+  swapExactTokensForETH = 'swapExactTokensForETH',
```
shresthagrawal, line 30: We don't use this I think

src/dex/kyberdmm.ts
nduchak: hmm...I thought that we always have one pool, but anyway this should work

shresthagrawal (approved):

PR #36, opened 2022-01-10
Don't use attach to avoid creation of new contract on every call

shresthagrawal:


src/dex/simple-exchange.ts
```diff
@@ -21,7 +20,7 @@ export class SimpleExchange {
   ) {
     this.simpleSwapHelper = new Interface(SimpleSwapHelperABI);
     // The contract address is set to null address as the token address is not known upfront
```
ColonelJ, line 22: This comment no longer makes sense and can be removed.

PR #43, opened 2022-02-07
Add support for new fee model

shresthagrawal:
Resolves BACK-378

### DIFF
--- a/src/router/payload-encoder.ts
+++ b/src/router/payload-encoder.ts
@@ -12,6 +12,23 @@ import { SwapSide } from '../constants';
 import { DexAdapterService } from '../dex';
 import { convertToBasisPoints } from '../utils';
 
+export function encodeFeePercent(
+  partnerFeePercent: string,
+  positiveSlippageToUser: boolean,
+) {
+  let fee = BigInt(partnerFeePercent);
+  if (fee > 10000) throw new Error('fee bps should be less than 10000');
+
+  // Set 14th bit if positiveSlippageToUser is true
+  if (positiveSlippageToUser) fee = fee | (BigInt(1) << BigInt(14));
+
+  // Bits 248 - 255 is used for version;
+  // Set version = 1;
+  fee = fee | (BigInt(1) << BigInt(248));
+
+  return fee.toString();
+}
+
 // This class can be used commonly by all the router
 // that will use the adapters.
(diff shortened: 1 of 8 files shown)


ColonelJ (changes requested):
There is a router for `buy` as well, not sure if you pulled it before starting this.

src/router/payload-encoder.ts
```diff
@@ -12,6 +12,23 @@ import { SwapSide } from '../constants';
...
+) {
+  let fee = BigInt(partnerFeePercent);
+  if (fee > 10000) throw new Error('fee bps should be less than 10000');
+
+  // Set 14th bit if positiveSlippageToUser is true
+  if (positiveSlippageToUser) fee = fee | (BigInt(1) << BigInt(14));
+
+  // Bits 248 - 255 is used for version;
+  // Set version = 1;
+  fee = fee | (BigInt(1) << BigInt(248));
```
ColonelJ, line 27: You can use `|=` on these two lines.

src/router/payload-encoder.ts
```diff
@@ -12,6 +12,23 @@ import { SwapSide } from '../constants';
...
 
+export function encodeFeePercent(
+  partnerFeePercent: string,
+  positiveSlippageToUser: boolean,
+) {
+  let fee = BigInt(partnerFeePercent);
+  if (fee > 10000) throw new Error('fee bps should be less than 10000');
+
+  // Set 14th bit if positiveSlippageToUser is true
+  if (positiveSlippageToUser) fee = fee | (BigInt(1) << BigInt(14));
```
erez-paraswap: I'd recommend calculating (BigInt(1) << BigInt(14) once and the using the value as much as needed

src/router/payload-encoder.ts
erez-paraswap: Same as above here

erez-paraswap (approved):
These are not functional problems, but we should definitely put more attention to code performance as we learn on our own skin

src/router/payload-encoder.ts
shresthagrawal: fixed

src/router/payload-encoder.ts
shresthagrawal: fixed

PR #51, opened 2022-02-28
Balancer/new pool types

johngrantuk:
* Added new PhantomStable and Linear pool types with associated maths and onchain data calls.
* Added checkBalance functions for each pool type. E.g. weighted pools have 30% swap limit but stable pools do not have this.
* Added extra tokens to runDexIntegrations.ts for easier testing of different pool types.
* Changed BasePool to use mul/divFixed versions to maintain same scaling as SC - different pool types were using different priceRate scaling i.e. Weighted would use 1 for priceRate but meta would use 1e18. This leads to larger number of decimal places after scaling and result was inaccurate compared to EVM maths.

### DIFF
--- a/src/dex/balancer-v2/LinearPool.ts
+++ b/src/dex/balancer-v2/LinearPool.ts
@@ -5,6 +5,7 @@ import * as LinearMath from './LinearMath';
 import { BZERO } from './balancer-v2-math';
 import { BasePool } from './balancer-v2-pool';
 import { callData, SubgraphPoolBase, PoolState, TokenState } from './types';
+import LinearPoolABI from '../../abi/balancer-v2/linearPoolAbi.json';
 
 export enum PairTypes {
   BptToMainToken,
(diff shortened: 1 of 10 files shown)


src/dex/balancer-v2/balancer-v2.ts
```diff
@@ -109,6 +113,8 @@ export class BalancerV2EventPool extends StatefulEventSubscriber<PoolStateMap> {
     'Weighted',
     'LiquidityBootstrapping',
     'Investment',
+    'StablePhantom',
+    'AaveLinear',
```
shresthagrawal, line 117: Both of the pools depend on external contract for rate similar to metaStable pools. Hence just tracking balances in events should not be enough. You can disable them from event system by removing them from the list. Hence the state will always be fetched on chain. This can be further improved by tracking the events of the rate contracts so that a complete event system can be built.

src/dex/balancer-v2/balancer-v2.ts
```diff
@@ -404,6 +497,20 @@ export class BalancerV2EventPool extends StatefulEventSubscriber<PoolStateMap> {
...
+            this.poolInterfaces,
+            this.vaultInterface,
+            data,
+            i,
+          );
+          i = newIndex;
+          acc = { ...acc, ...decoded };
+          return acc;
+        }
+
```
shresthagrawal, line 513: May be it would be better to refactor the the getOnChainState logic for all types of pools to have `getOnChainCalls`, `decodeOnChainCalls` functions in the pool class similar to what you did for linear pools. Also the poolInterface can be simply be the member of the poolClass so the interface is nor required by the `BalancerV2EventPool`.

src/dex/balancer-v2/LinearPool.ts
```diff
@@ -0,0 +1,380 @@
...
+      callData: poolInterface.encodeFunctionData('getWrappedIndex'),
+    });
+    poolCallData.push({
+      target: pool.address,
+      callData: poolInterface.encodeFunctionData('getBptIndex'),
+    });
+    // returns lowerTarget, upperTarget
+    poolCallData.push({
+      target: pool.address,
+      callData: poolInterface.encodeFunctionData('getTargets'),
```
shresthagrawal, line 281: It is recommended to only query state which changes. Params like `getMainIndex`, `getWrappedIndex`, `getBptIndex` are fixed and can be queried in the initial subgraph query

src/dex/balancer-v2/balancer-v2.ts
johngrantuk: I've refactored as you suggested and also added a parsePoolPairData helper for each pool type as I thought it helped keep things more organised.

PR #56, opened 2022-03-08
Feature/back 437 - create tests for UniswapV2 forks

Verisana:


src/dex/uniswap-v2/uniswap-v2-e2e-mainnet.test.ts
```diff
@@ -0,0 +1,1333 @@
...
+      provider,
+    );
+  });
+
+  describe('UniswapV2', () => {
+    const dexKey = 'UniswapV2';
+    const network = Network.MAINNET;
+    const tokens = Tokens[Network.MAINNET];
+    const holders = Holders[Network.MAINNET];
+    const provider = new JsonRpcProvider(ProviderURL[network]);
```
ColonelJ, line 39: don't need these 4 lines, the values are already set above...

ColonelJ (changes requested):
There are a lot of tests in the mainnet UniswapV2 E2E which don't have a dexKey and therefore don't belong in dex-lib (can use any DEX).  I think they should be left in API's e2e tests, and removed from here.

src/dex/uniswap-v2/uniswap-v2-e2e-mainnet.test.ts
Verisana: Fixed

Verisana:
@ColonelJ 
I agree. Removed from this tests all blocks on `Mainnet`:

```
ContractMethod: BuyMethod
ContractMethod: MegaPath
Multiswap Edge Cases
```
And two test cases on `Polygon` and `Mainnet` without clear `dexKey`

ColonelJ (approved):

PR #64, opened 2022-03-17
Balancer/new pool types

johngrantuk:
* Added new PhantomStable and Linear pool types with associated maths and onchain data calls.
* Added checkBalance functions for each pool type. E.g. weighted pools have 30% swap limit but stable pools do not have this.
* Added extra tokens to runDexIntegrations.ts for easier testing of different pool types.
* Changed BasePool to use mul/divFixed versions to maintain same scaling as SC - different pool types were using different priceRate scaling i.e. Weighted would use 1 for priceRate but meta would use 1e18. This leads to larger number of decimal places after scaling and result was inaccurate compared to EVM maths.

TO DO - Add testing.


src/dex/balancer-v2/balancer-v2-pool.ts
```diff
@@ -414,4 +577,142 @@ export class WeightedPool extends BaseMinimalSwapInfoPool {
...
+  data must contain returnData
+  startIndex is where to start in returnData. Allows this decode function to be called along with other pool types.
+  */
+  static decodeOnChainCalls(
+    pool: SubgraphPoolBase,
+    vaultInterface: Interface,
+    data: { success: boolean; returnData: any }[],
+    startIndex: number,
+  ): [{ [address: string]: PoolState }, number] {
+    const poolInterface = new Interface(WeightedPoolABI);
```
shresthagrawal, line 653: Interfaces are expensive to create and should be created in the constructor! I would suggest making decodeOnChainCalls and getOnChainCalls just normal members of the pool class (not static). You can create the pool interface in the constructor. Also you can pass vaultInterface in the constructor instead of passing that in each function call

src/dex/balancer-v2/balancer-v2-pool.ts
```diff
@@ -414,4 +577,142 @@ export class WeightedPool extends BaseMinimalSwapInfoPool {
...
+
+  /*
+  Helper function to construct onchain multicall data for WeightedPool (also 'LiquidityBootstrapping' and 'Investment' pool types).
+  */
+  static getOnChainCalls(
+    pool: SubgraphPoolBase,
+    vaultAddress: string,
+    vaultInterface: Interface,
+  ): callData[] {
+    const poolInterface = new Interface(WeightedPoolABI);
```
shresthagrawal, line 625: Interfaces are expensive to create and should be created in the constructor! See the comment below for more details on fix

src/dex/balancer-v2/balancer-v2.ts
```diff
@@ -328,50 +372,51 @@ export class BalancerV2EventPool extends StatefulEventSubscriber<PoolStateMap> {
...
-          poolCallData.push({
-            target: pool.address,
-            callData:
-              this.poolInterfaces['MetaStable'].encodeFunctionData(
-                'getScalingFactors',
-              ),
-          });
-        }
+        let poolCallData = [];
         if (
```
shresthagrawal, line 376: Though the code is logically correct. 
It would be better to use `switch` instead of `if if if ..` because a pooltype will only call one `getOnChainCalls`. Same structure for decode. Another way would be just use an dict to store a map from poolType to poolImplementation in the constructor. Then simply do `dict[poolType].getOnChainCalls(...)/decodeOnChainCalls(...)`

src/dex/balancer-v2/LinearMath.ts
```diff
@@ -0,0 +1,299 @@
...
+  }
+
+  const previousNominalMain = _toNominal(mainBalance, params);
+  const afterNominalMain = _toNominal(mainBalance + mainIn, params);
+  const deltaNominalMain = afterNominalMain - previousNominalMain;
+  const invariant = _calcInvariantUp(previousNominalMain, wrappedBalance);
+  return MathSol.divDownFixed(
+    MathSol.mulDownFixed(bptSupply, deltaNominalMain),
+    invariant,
+  );
```
shresthagrawal, line 29: I am checking against the implementation https://github.com/balancer-labs/balancer-v2-monorepo/blob/master/pkg/pool-linear/contracts/LinearMath.sol#L75
Shouldn't this be? 
```suggestion
  return MathSol.divDown(
    MathSol.mul(bptSupply, deltaNominalMain),
    invariant,
  );
```

shresthagrawal:
I started checking the Math for `src/dex/balancer-v2/LinearPool.ts` and comparing it with [Contract Implementation](https://github.com/balancer-labs/balancer-v2-monorepo/blob/master/pkg/pool-linear/contracts/LinearMath.sol). I see a lot of discrepancy in operators used. Also you have used fixed point calculation everywhere but the contract uses the normal operations. Are you using something other than contracts as the source of truth? I would suggest to follow exactly like the contracts so that its easier to verify it.

src/dex/balancer-v2/balancer-v2-pool.ts
johngrantuk: Didn't realise Interfaces are so expensive! Changed as suggested.

src/dex/balancer-v2/balancer-v2.ts
johngrantuk: Refactored and also applied same logic to `getPricesPool`.

johngrantuk:
> I started checking the Math for `src/dex/balancer-v2/LinearPool.ts` and comparing it with [Contract Implementation](https://github.com/balancer-labs/balancer-v2-monorepo/blob/master/pkg/pool-linear/contracts/LinearMath.sol). I see a lot of discrepancy in operators used. Also you have used fixed point calculation everywhere but the contract uses the normal operations. Are you using something other than contracts as the source of truth? I would suggest to follow exactly like the contracts so that its easier to verify it.

I'll check this with Sergio who implemented the maths and get back to you.

src/dex/balancer-v2/LinearMath.ts
```diff
@@ -0,0 +1,285 @@
...
+}
+
+export function _calcWrappedInPerMainOut(
+  mainOut: bigint,
+  mainBalance: bigint,
+  params: Params,
+): bigint {
+  // Amount in, so we round up overall.
+  const previousNominalMain = _toNominal(mainBalance, params);
+  const afterNominalMain = _toNominal(mainBalance - mainOut, params);
```
shresthagrawal, line 88:
```suggestion
  const afterNominalMain = _toNominal(mainBalance + mainOut, params);
```
As per https://github.com/balancer-labs/balancer-v2-monorepo/blob/master/pkg/pool-linear/contracts/LinearMath.sol#L102

src/dex/balancer-v2/LinearPool.ts
```diff
@@ -0,0 +1,372 @@
...
+    );
+  }
+
+  /*
+     Called when a swap with the Pool occurs, where the amount of tokens entering the Pool is known.
+     All amounts are upscaled.
+     Swap fee is NOT already deducted.
+     The return value is also considered upscaled, and should be downscaled (rounding down)
+     */
+  _onSwapGivenIn(
```
shresthagrawal, line 153: Is this function complete? I am not sure if all the cases are handled as per the contract https://github.com/balancer-labs/balancer-v2-monorepo/blob/master/pkg/pool-linear/contracts/LinearPool.sol#L253

shresthagrawal: We have 6 pair types but we are only handling 2?

shresthagrawal (approved):

ColonelJ (approved):

PR #69, opened 2022-03-30
excalibur integration

percivalExcalibur:
Excalibur is forked from Uniswapv2 with dynamic swap fees, deployed on Fantom
Factory contract:  https://ftmscan.com/address/0x08b3CCa975a82cFA6f912E0eeDdE53A629770D3f

### DIFF
--- a/src/dex/uniswap-v2/excalibur.ts
+++ b/src/dex/uniswap-v2/excalibur.ts
@@ -48,11 +48,13 @@ export class Excalibur extends UniswapV2 {
       callData: this.excaliburPool.encodeFunctionData('feeAmount', []),
     };
     const callDecoder = (values: any[]) =>
-      parseInt(
-        this.excaliburPool
-          .decodeFunctionResult('feeAmount', values)[0]
-          .toString(),
-      ) / 10;
+      Math.ceil(
+        parseInt(
+          this.excaliburPool
+            .decodeFunctionResult('feeAmount', values)[0]
+            .toString(),
+        ) / 10,
+      );
     return {
       callEntry,
       callDecoder,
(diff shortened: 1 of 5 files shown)


src/dex/uniswap-v2/excalibur.ts
```diff
@@ -0,0 +1,62 @@
...
+      network,
+      dexKey,
+      dexHelper,
+      true,
+      ExcaliburConfig[dexKey][network].factoryAddress,
+      ExcaliburConfig[dexKey][network].subgraphURL,
+      ExcaliburConfig[dexKey][network].initCode,
+      ExcaliburConfig[dexKey][network].feeCode,
+    );
+    this.feeFactor = 100000;
```
ColonelJ, line 42: Unfortunately, it's not valid to change feeFactor from 10000, because our smart contracts assume 10000.  Not sure why we left it possible to overwrite it like this.  The tests might still pass, but 10x fees will be charged on the swap in our router, which is obviously bad.  You will have to instead divide the fee by 10 (rounding up to next integer) in the callDecoder (prices will be marginally worse when fee is not a multiple of 10).  You can see in Biswap that we did the opposite to the fee (multiply by 10).

ColonelJ (changes requested):

percivalExcalibur:
Thanks for the clarification!
It should be fixed now

src/dex/uniswap-v2/excalibur.ts
```diff
@@ -0,0 +1,61 @@
...
+    const callEntry = {
+      target: poolAddress,
+      callData: this.excaliburPool.encodeFunctionData('feeAmount', []),
+    };
+    const callDecoder = (values: any[]) =>
+      parseInt(
+        this.excaliburPool
+          .decodeFunctionResult('feeAmount', values)[0]
+          .toString(),
+      ) / 10;
```
ColonelJ, line 55: You need to use Math.ceil() on the result of dividing by 10 to ensure you have an integer value.

ColonelJ (changes requested):

ColonelJ (dismissed):

shresthagrawal (approved):
LGTM! Ran the tests and everything is passing

ColonelJ (approved):

PR #70, opened 2022-03-31
Feature/back 466 - integrate new dex `nerve` and forks with event base system

Verisana:


src/dex/nerve/config.ts
```diff
@@ -0,0 +1,467 @@
+import { SwapSide } from 'paraswap-core';
+import type { AbiItem } from 'web3-utils';
+import { AdapterMappings, DexParams } from './types';
+import { DexConfigMap } from '../../types';
+import { Network } from '../../constants';
+const nervePoolABI = require('../../abi/nerve/nerve-pool.json');
+const axialPoolABI = require('../../abi/nerve/axial-pool.json');
+const ironV2PoolABI = require('../../abi/nerve/iron-v2-pool.json');
```
shresthagrawal, line 8: why not use import <name> from '<abi-location>'?

src/dex/nerve/config.ts
```diff
@@ -0,0 +1,467 @@
...
+        //     '0x07663837218A003e66310a01596af4bf4e44623D', // 0 - rUSD
+        //     '0xf2511b5E4FB0e5E2d123004b672BA14850478C14', // 1 - 3Pool-LP
+        //   ],
+        //   isMetapool: true,
+        //   lpTokenAddress: '0x870ee4d19c12A789c61de69E3E5eFb42383E4434',
+        // },
+      },
+    },
+  },
+
```
shresthagrawal, line 114:
```suggestion
```

src/dex/nerve/config.ts
```diff
@@ -0,0 +1,467 @@
...
+  },
+};
+
+export const dexKeyToABIMap: Record<string, AbiItem[]> = {
+  Nerve: nervePoolABI,
+  Axial: axialPoolABI,
+  IronV2: ironV2PoolABI,
+  // The same as axial
+  Saddle: axialPoolABI,
+};
```
shresthagrawal, line 449: You can also put this in the DexConfigMap. Also did you check if the ABI is different? If they are forks the ABI might be exactly same

src/dex/nerve/config.ts
Verisana: I just wanted to get autocompletion for functions on Contracts, but it didn't work. Replaced with ordinary `import <name> from '<abi-location>'`

src/dex/nerve/config.ts
Verisana: Fixed

src/dex/nerve/config.ts
Verisana: Yes, put the `abi` property to config. Seems good

src/dex/nerve/getstate-multicall.ts
```diff
@@ -0,0 +1,124 @@
...
+    };
+    return (pool as NerveEventMetapool).basePool
+      ? {
+          ..._state,
+          baseVirtualPrice: BigInt('0'),
+          baseCacheLastUpdated: BigInt('0'),
+        }
+      : _state;
+  });
+}
```
shresthagrawal, line 78: This doesn't look correct at all to me! you are fetching so many things but you are only decoding the `swapStorage`!

Verisana: Sorry for misleading you. As we discussed in DMs, this part is not used anymore. Deleted it.

src/dex/nerve/nerve-math.ts
```diff
@@ -0,0 +1,317 @@
...
+      tokenIndexTo,
+      x,
+      xp,
+      blockTimestamp,
+    );
+
+    if (y === null) {
+      state.isValid = false;
+      return null;
+    }
```
shresthagrawal, line 39: When do you get an invalid state?

src/dex/nerve/nerve-math.ts
```diff
@@ -0,0 +1,317 @@
...
+    // dy = dy.mul(FEE_DENOMINATOR.sub(calculateCurrentWithdrawFee(self, account)))
+    //  .div(FEE_DENOMINATOR);
+    dy =
+      (dy * (this.FEE_DENOMINATOR - this._calculateCurrentWithdrawFee(state))) /
+      this.FEE_DENOMINATOR;
+
+    return { dy, dyFee: dySwapFee };
+  }
+
+  swapUnderlying(state: PoolState) {}
```
shresthagrawal, line 83: is it suppose to be here?

src/dex/nerve/nerve-math.ts
```diff
@@ -0,0 +1,317 @@
...
+    // v.d1 = v.d0.sub(tokenAmount.mul(v.d0).div(self.lpToken.totalSupply()));
+    v.d1 = v.d0 - (tokenAmount * v.d0) / state.lpToken_supply;
+    const newY = this._getYD(state, v.preciseA, tokenIndex, xp, v.d1);
+    if (newY === null) return null;
+    else v.newY = newY;
+
+    const xpReduced: bigint[] = [];
+
+    v.feePerToken = this._feePerToken(state);
+    for (let i = 0; i < numTokens; i++) {
```
shresthagrawal, line 118: Does this work? Isn't `numTokens` a bigint while i is a `number`

src/dex/nerve/nerve-math.ts
```diff
@@ -0,0 +1,317 @@
...
+      const xpi = xp[i];
+      // if i == tokenIndex, dxExpected = xp[i] * d1 / d0 - newY
+      // else dxExpected = xp[i] - (xp[i] * d1 / d0)
+      // xpReduced[i] -= dxExpected * fee / FEE_DENOMINATOR
+
+      const dxExpected =
+        i === tokenIndex
+          ? (xpi * v.d1) / v.d0 - v.newY
+          : xpi - (xpi * v.d1) / v.d0;
+      xpReduced[i] -= (dxExpected * v.feePerToken) / this.FEE_DENOMINATOR;
```
shresthagrawal, line 128: according to the nerve contract this is rather
```suggestion
      const dxExpected =
        i === tokenIndex
          ? (xpi * v.d1) / v.d0 - v.newY
          : xpi - (xpi * v.d1) / v.d0;
      xpReduced[i] = xpi - (dxExpected * v.feePerToken) / this.FEE_DENOMINATOR;
```

src/dex/nerve/nerve-math.ts
```diff
@@ -0,0 +1,317 @@
...
+    const numTokens = this._getNumTokens(state);
+    // self.swapFee.mul(self.pooledTokens.length).div(self.pooledTokens.length.sub(1).mul(4));
+    return (state.swapFee * numTokens) / ((numTokens - ONE) * biginterify(4));
+  }
+
+  protected _calculateCurrentWithdrawFee(state: PoolState) {
+    // It is not correct. We should calculate user withdrawFeeMultiplier by
+    // the time passed since the liquidity was added
+    return state.defaultWithdrawFee ? state.defaultWithdrawFee : ZERO;
+  }
```
shresthagrawal, line 152: how is it handled when the withdraw event is released?

src/dex/nerve/nerve-math.ts
```diff
@@ -0,0 +1,317 @@
...
+      // y = y.mul(y).add(c).div(y.mul(2).add(b).sub(d));
+      y = (y * y + c) / (y * biginterify(2) + b - d);
+      if (MathUtil.within1(y, yPrev)) {
+        return y;
+      }
+    }
+
+    this.logger.error(`Event pool ${this.name} method _getYD did not converge`);
+    state.isValid = false;
+    return null;
```
shresthagrawal, line 196: may be its better to throw error instead of logging and returning null. This will make the implementation more clean and also similar to the contract. You can have a try catch in the outer level to handle it

src/dex/nerve/nerve-math.ts
```diff
@@ -0,0 +1,317 @@
...
+      if (MathUtil.within1(y, yPrev)) {
+        return y;
+      }
+    }
+
+    this.logger.error(
+      `Event pool ${this.name} parsing function _getY approximation did not converge`,
+    );
+    state.isValid = false;
+    return null;
```
shresthagrawal, line 250: same as before

src/dex/nerve/nerve-math.ts
```diff
@@ -0,0 +1,317 @@
...
+      if (MathUtil.within1(d, prevD)) {
+        return d;
+      }
+    }
+
+    // Convergence should occur in 4 loops or less. If this is reached, there may be something wrong
+    // with the pool.
+    this.logger.error(`Event pool ${this.name} method _getD did not converge`);
+    state.isValid = false;
+    return null;
```
shresthagrawal, line 309: same as above

src/dex/nerve/nerve-math.ts
shresthagrawal: this somehow magically works

src/dex/nerve/nerve-math.ts
shresthagrawal: Check the comments below on better handling of this

src/dex/nerve/nerve-math.ts
Verisana: Its a small step in the future when we support meta pools :)
Removed

src/dex/nerve/nerve-math.ts
Verisana: Pure lucky on my side that it actually works.

> A Number value and a BigInt value may be compared as usual:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt

src/dex/nerve/nerve-math.ts
Verisana: Actually, we don't use this as its not fully correct. I just invalidate the state, so it can be updated by RPC call

src/dex/nerve/nerve-math.ts
Verisana: Thanks, it looks much better. I didn't like this `null` checks. And I am glad to get rid of them.

src/dex/nerve/nerve-math.ts
Verisana: We still have invalid state for unsupported states and when error occured while calculating new state, but it become more readable

src/dex/nerve/nerve-pool.ts
```diff
@@ -0,0 +1,433 @@
...
+          `For ${this.parentName}_${this.poolConfig.name} _calculateSwap value ${dy} is not equal to ${dyEvent} event value`,
+        );
+        state.isValid = false;
+        return null;
+      }
+
+      return state;
+    } catch (e) {
+      state.isValid = false;
+      return null;
```
ColonelJ, line 308: You can't return `null` if you need to change the state (by setting `isValid` to `false` for example) because then the change won't be processed.  Return `state` instead.  (Happens in a lot of places.)

src/dex/nerve/nerve.ts
```diff
@@ -0,0 +1,399 @@
...
+          const out = pool.math.calculateSwap(
+            _state,
+            srcIndex,
+            destIndex,
+            _amount,
+            // Actually we need here block timestamp, but +- 15 seconds shouldn't
+            // affect the calculations
+            bigIntify((Date.now() / 1000).toFixed(0)),
+          );
+          if (out === null) {
```
ColonelJ, line 219: I don't think `calculateSwap` returns `null` any more?  So you should catch the error.

src/dex/nerve/nerve.ts
```diff
@@ -0,0 +1,405 @@
...
+      // here side === SwapSide.SELL
+      const unitVolume = getBigIntPow(10 ** _srcToken.decimals);
+      const chunks = amounts.length - 1;
+
+      const _width = Math.floor(chunks / NERVE_CHUNKS);
+      const _amounts = [unitVolume].concat(
+        Array.from(Array(NERVE_CHUNKS).keys()).map(
+          i => amounts[(i + 1) * _width],
+        ),
+      );
```
shresthagrawal, line 197: Do we need to use lower chunks for Nerve? It should be fine to just to pricing for all the volumes

src/dex/nerve/nerve.ts
```diff
@@ -0,0 +1,405 @@
...
+  async getStates(
+    pools?: EventPoolOrMetapool[],
+    blockNumber?: number,
+  ): Promise<DeepReadonly<{ state: PoolState; pool: EventPoolOrMetapool }[]>> {
+    const _pools = pools === undefined ? this.allPools : pools;
+
+    const _blockNumber =
+      blockNumber === undefined
+        ? await this.dexHelper.provider.getBlockNumber()
+        : blockNumber;
```
shresthagrawal, line 110: if the blocknumber is not specified it is assumed it should be latest. In that case you don't need to an extra rpc call here to resolve the blocknumber. generate state already accepts `latest` as blocknumber

src/dex/nerve/nerve.ts
```diff
@@ -0,0 +1,405 @@
...
+            `State for ${this.dexKey} pool ${eventPool.name} is stale or invalid on block ${_blockNumber}. Generating new one`,
+          );
+          const newState = await eventPool.generateState(_blockNumber);
+          eventPool.setState(newState, _blockNumber);
+          return { state: newState, pool: eventPool };
+        } else {
+          return { state, pool: eventPool };
+        }
+      }),
+    );
```
shresthagrawal, line 126: This will do multiple RPC requests! Technically this should be batched into single RPC. May be its fine to go with this for now as there are not many pool but at least mention a `TODO:` comment that can be fixed in future

src/dex/nerve/nerve.ts
```diff
@@ -0,0 +1,405 @@
...
+  }
+
+  async getPoolIdentifiers(
```
shresthagrawal, line 129: the `getPriceVolume` does wrapETH so the `getPoolIdentifier` should also do the same!

src/dex/nerve/nerve.ts
ColonelJ: A block number is needed for getState and setState...

src/dex/nerve/nerve.ts
shresthagrawal: Usually lower chunks are required if you RPC or if the computation is very heavy. With event systems its usually better to get prices for all the volumes

src/dex/nerve/types.ts
```diff
@@ -0,0 +1,81 @@
...
+  futureATime: bigint;
+  swapFee: bigint;
+  adminFee: bigint;
+  defaultDepositFee?: bigint;
+  defaultWithdrawFee?: bigint;
+  lpToken_supply: bigint;
+  balances: bigint[];
+  tokenPrecisionMultipliers: bigint[];
+  isValid: boolean;
+}
```
shresthagrawal, line 21:
```suggestion
export type PoolState = {
  initialA: bigint;
  futureA: bigint;
  initialATime: bigint;
  futureATime: bigint;
  swapFee: bigint;
  adminFee: bigint;
  defaultDepositFee?: bigint;
  defaultWithdrawFee?: bigint;
  lpToken_supply: bigint;
  balances: bigint[];
  tokenPrecisionMultipliers: bigint[];
  isValid: true;
} | { isValid: false }
```
This can be used to avoid cloning the state

src/dex/nerve/nerve.ts
```diff
@@ -0,0 +1,405 @@
...
+      const _amounts = [unitVolume].concat(
+        Array.from(Array(NERVE_CHUNKS).keys()).map(
+          i => amounts[(i + 1) * _width],
+        ),
+      );
+
+      const result: ExchangePrices<NerveData> = [];
+      for (const { pool, state } of statePoolPair) {
+        // To be able to update state property isValid if calculations receive error
+        const _state = _.cloneDeep(state) as PoolState;
```
shresthagrawal, line 202: I don't think you need to clone the state! You simply set the invalid flag to true but you don't reuse the old state. I suggested a type change below for poolState

src/dex/nerve/nerve.ts
shresthagrawal: True

shresthagrawal:
If the metapools is not supported would it be better to remove the metapools code to avoid confusion?

src/dex/nerve/nerve.ts
ColonelJ: Right now, the NervePoolMath functions are allowed to modify the state, so cloning it makes sense, but I would suggest making those functions take a DeepReadonly<State> instead and not modify the state, only throw the exception, so that there is no need to clone, then the exception handlers can set the isValid flag to false (in pool event handlers, I think this is already done).

src/dex/nerve/nerve.ts
Verisana: I actually didn't understand this chunking scheme and reimplemented the one which was originally used. Thanks for clarifying :)
Fixed

src/dex/nerve/nerve.ts
Verisana: Fixed

src/dex/nerve/types.ts
Verisana: Fixed another way as we discussed

src/dex/nerve/nerve.ts
Verisana: Fixed another way as we discussed

shresthagrawal (approved):

ColonelJ (approved):

src/dex/nerve/nerve.ts
Verisana: For a while lets put there `TODO`

PR #72, opened 2022-04-04
add solidly and printy

printyfi:
https://github.com/solidlyexchange/solidly#mainnet
https://github.com/printyfi/printy-solidly-contracts#mainnet-red-chain-avalanche


shresthagrawal:
@printyfi Thanks for the work but neither Solidly nor Printify seems to be a UniswapV2 fork. Hence you can just append the config to the fork. If you would like to integrate those DEXes you would have to create a custom integration. You should check the [README](https://github.com/paraswap/paraswap-dex-lib/blob/master/README.md) for more details

PR #86, opened 2022-04-09
printy.finance as uniswapv2 swap testing 

printyfi:
since the factory contract is very similar to uniswapv2  it might work for price quotes and swaps ?

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@paraswap/dex-lib",
-  "version": "2.6.1",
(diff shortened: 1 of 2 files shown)


shresthagrawal:
Its not enough to have just the factory same! The pricing and events logic should be the same as well

printyfi:
> Its not enough to have just the factory same! The pricing and events logic should be the same as well

than please give me an easy step by step guide , this is to complicated .

avadex:
@shresthagrawal

PR #90, opened 2022-04-18
TetuSwap Integration

bogdoslavik:
TetuSwap - the part of TETU https://tetu.io/
Docs: https://docs.tetu.io/tetu-io/tetuswap/tetuswap
Addresses: https://docs.tetu.io/tetu-io/protocol/addresses


bogdoslavik:
I have one question:
TetuSwap is UniswapV2 fork, but we have different initCode for different pairs. 

All default e2e tests passed (simpleSwap, multiSwap, megaSwap).

Do I have to provide initCode for each pair (that used in buyOnUniswapFork method only), or it is not necessary?

src/dex/uniswap-v2/tetuswap.ts
```diff
@@ -0,0 +1,120 @@
...
+    [Network.POLYGON]: {
+      subgraphURL: 'https://api.thegraph.com/subgraphs/name/tetu-io/tetu-swap',
+      factoryAddress: '0x684d8c187be836171a1Af8D533e4724893031828',
+      router: '0xBCA055F25c3670fE0b1463e8d470585Fe15Ca819',
+      // To PR reviewer:
+      // TetuSwap init code is different for different pairs
+      // does it need additional configuration to use buyOnUniswapV2Fork only
+      // and do not use buyOnUniswapFork method that needs initCode?
+      initCode: '0x0',
+      poolGasCost: 1000 * 1000, // TetuSwap use SmartVault deposits / withdrawals during swap, so its costly
```
shresthagrawal, line 24: Is it really `1 million` cost?

src/dex/uniswap-v2/tetuswap.ts
```diff
@@ -0,0 +1,120 @@
...
+      feeCode: 10,
```
shresthagrawal, line 25: If its dynamic its better to set the feeCode to zero and put a comment

shresthagrawal:
In the #92 we are removing the `swap/buyOnUniswapFork` methods as they are legacy. This will also remove the requirement of init code

shresthagrawal:
As you have a special function to get reserves you should also override the `generateState` ([here](https://github.com/tetu-io/paraswap-dex-lib/blob/master/src/dex/uniswap-v2/uniswap-v2.ts#L138)) in the pool implementation for tetuswap. I think you should extend the `UniswapV2EventPool` to `TetuEventPool` with the correct function implementation. For using it with the `TetuSwap` class you also need either override the `addPool` function or simply fix the base `UniswapV2` implementation to accept the event pool implementation as the last constructor parameter with the default as the `UniswapV2EventPool`

src/dex/uniswap-v2/tetuswap.ts
bogdoslavik: it different, but about 1million  - because it invest / withdraws from yeld farming strategies under the hood. See router's swap transactions https://polygonscan.com/address/0xbca055f25c3670fe0b1463e8d470585fe15ca819

src/dex/uniswap-v2/tetuswap.ts
bogdoslavik: ok

bogdoslavik:
> As tetu pools have a special function to get reserves

Original uniswapV2 have [getReserves](https://github.com/Uniswap/v2-core/blob/4dd59067c76dea4a0e8e4bfdda41877a6b16dedc/contracts/UniswapV2Pair.sol#L38) function too. 

I am little surprised what you gets token balances twice, instead of using standard function that gets reserves from private variables.

You may get some swap errors (like `UniswapV2: K`) when reserves not in sync - then your code use token balances, while uniswap v2 logic uses reserves. 

But it just my opinion.

Anyway, I going to extend `UniswapV2EventPool` to `TetuEventPool` as you suggested.

shresthagrawal:
@bogdoslavik thats a good point. The original implementation was done by @ColonelJ. He confirmed it would be fine to switch the base implementation to use `getReserves` as well. If would be really amazing if you can fix the base implementation and then you don't have to do all the patching

bogdoslavik:
> He confirmed it would be fine to switch the base implementation to use `getReserves` as well. If would be really amazing if you can fix the base implementation and then you don't have to do all the patching

Great! Will fix the base implementation. It will make things easier.

One more question: I do not see any generateState calls from the tests. How I can test this function in right way?

bogdoslavik:
Hello again! :) 
I've updateв uniswapv2 base to use getReserves instead of balances.

Please review and test.

PR #93, opened 2022-04-20
Revert UniswapV2 cleanup

ColonelJ:
Causing higher error rate on API and also Dfyn WETH handling was broken.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@paraswap/dex-lib",
-  "version": "2.0.18",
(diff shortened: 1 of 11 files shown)


Louis-Amas (approved):

ColonelJ:
The UniswapV2 optimizer wasn't updated, that could be where the errors are coming from (it uses the tokens `path`, which was removed, and probably shouldn't be).

erez-paraswap (approved):

PR #94, opened 2022-04-20
provider: use generic class

Louis-Amas:
- add network to prevent useless rpc call

Resolves BACK-526.


ColonelJ (changes requested):
You missed changing to `StaticJsonRpcProvider` in `dex-template` and `scripts` directories.

Louis-Amas:
> 

fix

ColonelJ (approved):

erez-paraswap (approved):
I wonder if we could got rid from ethersjs alltogether

PR #97, opened 2022-04-22
uniswap-v2: fix case

Louis-Amas:
Resolves BACK-533.


Verisana (approved):

shresthagrawal (approved):

ColonelJ (changes requested):
I think UniswapV2Alias also should be lowercased.

ColonelJ (approved):

PR #101, opened 2022-04-27
feat: ib-amm integration

wei3erHase:
Co-authored-by:  0xUwUnt<uwunt@defi.sucks>
Co-authored-by: 0xng <ng@defi.sucks>

### DIFF
--- a/src/dex/ib-amm/ib-amm.ts
+++ b/src/dex/ib-amm/ib-amm.ts
@@ -1,0 +1,215 @@
+import { Interface } from '@ethersproject/abi';
+import {
+  Token,
+  Address,
+  ExchangePrices,
+  AdapterExchangeParam,
+  SimpleExchangeParam,
+  PoolLiquidity,
+  Logger,
+} from '../../types';
+import { SwapSide, Network, ProviderURL } from '../../constants';
+import { getDexKeysWithNetwork } from '../../utils';
+import { IDex } from '../../dex/idex';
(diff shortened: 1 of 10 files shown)


wei3erHase:
@shresthagrawal @erez-paraswap 👀

src/dex/ib-amm/ib-amm.ts
```diff
@@ -0,0 +1,215 @@
...
+
+      if (this.config.IB_TOKENS.every(a => toLC(a) !== toLC(from.address)))
+        return false;
+    }
+
+    return true;
+  }
+
+  private getQuote(isBuy: boolean) {
+    const provider = new JsonRpcProvider(ProviderURL[Network.MAINNET]);
```
shresthagrawal, line 66: You should not create a provider on every quote call! You should reuse the provider available in the dexHelper

shresthagrawal:
The current implementation doesn't seem to be event based. Please read the documentation about event based systems [here](https://github.com/paraswap/paraswap-dex-lib#understanding-the-event-based-pricing-approach). I would advise to check out the existing event based dexes (like UniswapV2, BalancerV2) to get an understanding of the event based integration

wei3erHase:
@shresthagrawal we've added a pool state being updated with the oracle events!

apparently all tests are passing, we hope this is going on the right path,
let us know if there's anything else to pay attention to,
thanks a lot!!
WH

PR #102, opened 2022-04-28
Add GMX 

shresthagrawal:


src/dex/gmx/fast-price-feed.ts
```diff
@@ -0,0 +1,228 @@
...
+  public processLog(
+    state: DeepReadonly<FastPriceFeedState>,
+    log: Readonly<Log>,
+    blockHeader: Readonly<BlockHeader>,
+  ): DeepReadonly<FastPriceFeedState> | null {
+    try {
+      const parsed = FastPriceFeed.fastPriceEventsInterface.parseLog(log);
+      switch (parsed.name) {
+        case 'PriceUpdate': {
+          const _state: FastPriceFeedState = state;
```
ColonelJ, line 54: You need to make a deep copy if you're going to modify the state, otherwise you are invalidating the state history!  (It's DeepReadonly for a reason...)

src/dex/gmx/gmx.ts
```diff
@@ -0,0 +1,275 @@
...
+          destToken,
+          this.augustusAddress,
+        ]),
+      ],
+      values: ['0', '0'],
+      networkFee: '0',
+    };
+  }
+
+  async updatePoolState(): Promise<void> {
```
ColonelJ, line 196: I don't think pool tracker or pools helper actually calls this, so it's not going to work.

src/dex/gmx/usdg.ts
```diff
@@ -0,0 +1,82 @@
...
+  }
+
+  public processLog(
+    state: DeepReadonly<USDGState>,
+    log: Readonly<Log>,
+    blockHeader: Readonly<BlockHeader>,
+  ): DeepReadonly<USDGState> | null {
+    try {
+      const parsed = USDG.interface.parseLog(log);
+      const _state: USDGState = state;
```
ColonelJ, line 39: Same as other comment, you are modifying the state in the state history, you need to copy this or construct a new object.

src/dex/gmx/vault.ts
```diff
@@ -0,0 +1,289 @@
...
+  }
+
+  public processLog(
+    state: DeepReadonly<VaultState>,
+    log: Readonly<Log>,
+    blockHeader: Readonly<BlockHeader>,
+  ): AsyncOrSync<VaultState | null> {
+    try {
+      const parsed = Vault.interface.parseLog(log);
+      const _state: VaultState = state;
```
ColonelJ, line 265: Same as previous comment, need to deep clone probably.

ColonelJ (changes requested):

src/dex/gmx/gmx.ts
shresthagrawal: I think it would be better to fix the pool helper to allow for this. We can leave this as it is and fix the pools helper as a separate task

ColonelJ (approved):

erez-paraswap (approved):

PR #103, opened 2022-05-02
Balancer/virtual boosted pool

johngrantuk:
A VirtualBoostedPool is a combination of Linear pools (with underlying tokens of interest) nested in a PhantomStable pool e.g. bb-a-USD which connects underlying USDC/DAI/USDT. Vault BatchSwap architecture can be used for gas efficient swaps between the underlying tokens.

* Added new VirtualBoostedPool type. VirtualPools will be seen as separate pools:
    * poolType:VirtualBoosted
    * `pool.tokens` contains a list of the underlying tokens of interest, e.g. DAI/USDC/USDT for bbausd
    * id will be the ref ID (e.g. Phantom ID for bbausd) appended with `-virtualboostedpool`
    * address will be the ref address (e.g. address for bbausd) appended with `-virtualboostedpool`
    * Class has helper functions to fetch onchain state (for all boosted pools in group) and pricing (chained swaps)
* Optimizer: Remains unchanged. VirtualPool swaps should be treated the same as other swaps. VirtualPools will be created with -virtualboostedpool added to end of ID this acts as a flag to process correctly in `getBalancerParam` function. Added tests to confirm functions and flag future breaking changes.
* getBalancerParam: Changed to handle VirtualPools where swaps > 1 in length. Assets list will have more than tokenIn and tokenOut and so the relevant indices used in each swap must be correct, this is handled by `updateSwapAssets` function. Added tests.
* getBalancerParam - Changed to handle VirtualPools where swaps > 1 in length. Assets list will have more than tokenIn and tokenOut and so the relevant indices used in each swap must be correct, this is handled by updateSwapAssets function. Added tests.
* Added integration tests for bbausd VirtualPool.
* TO DO - Waiting for accurate gas costs. Adding accurate gas costs for each PoolType. Currently some failing tests.


src/dex/balancer-v2/balancer-v2-integration.test.ts
```diff
@@ -59,9 +66,9 @@ describe('BalancerV2', function () {
         pools,
       );
       console.log('WETH <> DAI Pool Prices: ', poolPrices);
-
       expect(poolPrices).not.toBeNull();
       checkPoolPrices(poolPrices!, amounts, SwapSide.SELL, dexKey);
+      expect(poolPrices?.[0].gasCost).toBe(25); // TO DO
```
johngrantuk, line 71: Currently failing but will be updated when gasCost is finalized.

src/dex/balancer-v2/balancer-v2-integration.test.ts
```diff
@@ -108,6 +115,7 @@ describe('BalancerV2', function () {
 
       expect(poolPrices).not.toBeNull();
       checkPoolPrices(poolPrices!, amounts, SwapSide.SELL, dexKey);
+      expect(poolPrices?.[0].gasCost).toBe(25); // TO DO
```
johngrantuk, line 118: Currently failing but will be updated when gasCost is finalized.

src/dex/balancer-v2/balancer-v2-integration.test.ts
```diff
@@ -176,4 +184,72 @@ describe('BalancerV2', function () {
...
+          'BalancerV2_0x7b50775383d3d6f0215a8f290f2c9e2eebbeceb2virtualboosted',
+      );
+      expect(virtualPoolIdentifier).not.toBeUndefined();
+      expect(virtualPoolIdentifier?.poolAddresses).toEqual([
+        '0x7b50775383d3d6f0215a8f290f2c9e2eebbeceb2virtualboosted',
+      ]);
+      expect(virtualPoolIdentifier?.data.poolId).toEqual(
+        '0x7b50775383d3d6f0215a8f290f2c9e2eebbeceb20000000000000000000000fevirtualboosted',
+      );
+      expect(virtualPoolIdentifier?.gasCost).toBe(25); // TO DO 2 * Linear + 1 * Phantom
```
johngrantuk, line 233: Currently failing but will be updated when gasCost is finalized.

src/dex/balancer-v2/balancer-v2.ts
```diff
@@ -659,19 +745,34 @@ export class BalancerV2
...
-    }`;
+            address
+            totalLiquidity
+            tokens {
+              address
+              decimals
+            }
+          }
+
+          poolsAll: pools(first: 1000, orderBy: totalLiquidity, orderDirection: desc, where: {swapEnabled: true, poolType_in: ["MetaStable", "Stable", "Weighted", "LiquidityBootstrapping", "Investment", "StablePhantom", "AaveLinear", "ERC4626Linear"]}) {
```
johngrantuk, line 762: VirtualBoostedPools creation needs info of all pools. VirtualBoostedPools could be taken from `this.eventPools.virtualBoostedPools` but in `balancer-v2-integration.test.ts` the `initializePricing` function isn't called so pools haven't been pre-loaded. Wasn't sure if this was a requirement.

src/dex/balancer-v2/VirtualBoostedPool.ts
```diff
@@ -0,0 +1,650 @@
...
+  vaultAddress: string;
+  vaultInterface: Interface;
+  linearPoolInterface: Interface;
+  phantomStablePoolInterface: Interface;
+
+  constructor(vaultAddress: string, vaultInterface: Interface) {
+    this.vaultAddress = vaultAddress;
+    this.vaultInterface = vaultInterface;
+    this.linearPoolInterface = new Interface(LinearPoolABI);
+    this.phantomStablePoolInterface = new Interface(MetaStablePoolABI);
```
shresthagrawal, line 131: It would be better to make these static members to avoid creating interface for each pool while it can be reused. This can also be fixed in other pool implementation

shresthagrawal (commented):
@johngrantuk I think you didn't understood the main idea that I described in the telegram chat. 
You need to fix the `getPoolIdentifiers` to return all the pools that will used when a virtual pool is used. This is important for our internal algorithm. So for example if X is boosted pool which underlying uses Y, Z then the getPoolIdentifiers should return X, Y, Z 

Similarly the getPricesVolume should check if X, Y, Z exist in the limit pool to even use X. The main reason is that the underlying pools can be used individually somewhere else in the route and our internal algorithm has to account for that to avoid collisions

src/dex/balancer-v2/balancer-v2.ts
shresthagrawal: Yes that is a requirement. But you can implement `updatePoolState` this is function that will be called  before `getTopPoolsForToken`. `updatePoolState` is called once before calling `getTopPoolsForToken` multiple times for different tokens. So you can fetch and save the state in the memory and reuse it in `getTopPoolsForToken`. I know this part is a bit confusing as you only know the opensource codebase but it is just a side effect of merging multiple dex interfaces into one

src/dex/balancer-v2/VirtualBoostedPool.ts
johngrantuk: Changed to create in main `BalancerV2EventPool` and reuse when instantiating pools.

src/dex/balancer-v2/balancer-v2.ts
johngrantuk: Implemented `updatePoolState` which fetches all pools from Subgraph, adds VirtualBoostedPools and stores in memory to be reused by `getTopPoolsForToken`

src/dex/balancer-v2/balancer-v2.ts
```diff
@@ -436,12 +497,33 @@ export class BalancerV2
...
+                  return false;
+              }
+              return true;
+            } else {
+              return limitPools.includes(
+                `${this.dexKey}_${pool.address.toLowerCase()}`,
+              );
+            }
+          })
+        : poolsWithTokens;
```
shresthagrawal, line 526: This doesn't seem correct! You should check for all the identifiers that you returned in `getPoolIdentidier` for the boosted pool. You are missing check for the main pool id and phantom pool id. 
Also if you use the boosted pools you must make sure you don't use the underlying pools.

src/dex/balancer-v2/balancer-v2.ts
```diff
@@ -482,7 +565,7 @@ export class BalancerV2
               _from,
               _to,
               pool,
-              poolState,
+              allPoolStates[poolAddress] ? allPoolStates : missingPoolsStateMap,
```
shresthagrawal, line 568: shouldn't you merge the `allPoolStates` and `missingPoolsStateMap`? what if some base pools are there is one of the object while  the others are there in other?

shresthagrawal (commented):
I don't think we need to append a superficial address with `${phantompooladdress}virtualboosted`. First of all, we expect pool addresses to be valid addresses. Second, it is sufficient to use linear pools plus phantom pools as pool addresses for the virtual boosted pool. If these pool addresses exist you can simply prioritize to use the boosted pool over the underlying pools

src/dex/balancer-v2/balancer-v2-integration.test.ts
shresthagrawal: for now I just added some constant based on the simulation, can be improved later

src/dex/balancer-v2/balancer-v2.ts
```diff
@@ -84,10 +94,16 @@ export class BalancerV2EventPool extends StatefulEventSubscriber<PoolStateMap> {
   } = {};
 
   pools: {
-    [type: string]: WeightedPool | StablePool | LinearPool | PhantomStablePool;
+    [type: string]:
+      | WeightedPool
+      | StablePool
+      | LinearPool
+      | PhantomStablePool
+      | VirtualBoostedPool;
```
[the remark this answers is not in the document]
shresthagrawal: this can be improved later

PR #110, opened 2022-05-17
Feature/back 585 - Support dynamic `gasCost` from `Dexlib`

Verisana:


src/pricing-helper.ts
```diff
@@ -172,6 +172,20 @@ export class PricingHelper {
...
         }
 
+        if (Array.isArray(p.gasCost)) {
+          if (p.gasCost.length !== amounts.length) {
+            this.logger.error(
+              `Error_getPoolPrices: ${p.exchange} returned prices with invalid gasCost array length: ${p.gasCost.length} !== ${amounts.length}`,
+            );
+            return false;
+          } else if (amounts[0] === 0n && p.gasCost[0] === 0) {
+            this.logger.error(
```
shresthagrawal, line 182: This condition doesn't seem correct

shresthagrawal: I think it should be
`if(amounts[0] === 0n && p.gasCost[0] !== 0)`

but may be its safer to check for every amount which is 0 if we don't want to make assumptions on the amounts

Verisana: Fixed

ColonelJ (approved):

PR #116, opened 2022-06-09
Maverick Integration

trqdinh:


src/dex/maverick/maverick-integration.test.ts
```diff
@@ -0,0 +1,75 @@
...
+      pools,
+    );
+    console.log(`${TokenASymbol} <> ${TokenBSymbol} Pool Prices: `, poolPrices);
+
+    expect(poolPrices).not.toBeNull();
+    if (maverick.hasConstantPriceLargeAmounts) {
+      checkConstantPoolPrices(poolPrices!, amounts, dexKey);
+    } else {
+      checkPoolPrices(poolPrices!, amounts, SwapSide.SELL, dexKey);
+    }
```
Verisana, line 58: Please add `checkOnChainPricing` from [here](https://github.com/paraswap/paraswap-dex-lib/blob/5ee32b9c33545d0fa2c8bde3726a9ccb2f191f76/dex-template/__DexName__-integration.test.ts.template#L146). It seems, we added it after you started the integration process:

But it is a good indicator if there is an error in math or not :)

src/dex/maverick/maverick-events.test.ts
```diff
@@ -0,0 +1,126 @@
...
+    { query, variables },
+    { timeout: 5000 },
+  );
+  return data.data.data.pools[0];
+}
+
+describe('Maverick Event', function () {
+  const blockNumbers: { [eventName: string]: number[] } = {
+    AddLiquidity: [29299844],
+    Swap: [29300959, 29301354, 29301378],
```
Verisana, line 76: Do you have more events? It would be good to include `RemoveLiquidity` too. As a rule of thumb we check at least 5 events if possible for each event

Verisana:
Hey @trqdinh,
Thanks a lot for your contribution! I run your tests, they pass. To make it work for other methods, I have to implement adapter for contract and check if it works. If everything is good, we can put it to production.

PR #117, opened 2022-06-10
integrate meshswap

sanjisong:
Adding Meshswap polygon AMM.
Meshswap is only compatible with uniswap v2 router, not pool.

App : https://meshswap.fi/
Docs: https://docs.meshswap.fi/


sanjisong:
And I have one question, do I have to write the part to add to the adapter code?

PR #120, opened 2022-06-10
Integrate Metavault.Trade 

0xrmvdao:


0xrmvdao:
We've updated Metavault.Trade integration and created new PR. no need to this one.

PR #121, opened 2022-06-14
feature/back-615 - `Dystopia` exchange integration

Verisana:
Integration from [PR#108](https://github.com/paraswap/paraswap-dex-lib/pull/108)


src/constants.ts
```diff
@@ -66,3 +66,5 @@ export const AugustusAddress: { [nid: number]: Address } = {
   [Network.AVALANCHE]: '0xDEF171Fe48CF0115B1d80b88dc8eAB59176FEe57',
   [Network.FANTOM]: '0xDEF171Fe48CF0115B1d80b88dc8eAB59176FEe57',
 };
+
+export const subgraphTimeout = 20 * 1000;
```
erez-paraswap, line 70: Special reason to have a constant starts in lowercase? Should be ALL_UPPERCASE

tests/constants-e2e.ts
```diff
@@ -535,7 +535,7 @@ export const Holders: {
   [Network.POLYGON]: {
     MATIC: '0xe7804c37c13166fF0b37F5aE0BB07A3aEbb6e245',
     DAI: '0x06959153B974D0D5fDfd87D561db6d8d4FA0bb0B',
-    WETH: '0xd3d176F7e4b43C70a68466949F6C64F06Ce75BB9',
+    WETH: '0x72a53cdbbcc1b9efa39c834a540550e23463aacb',
```
erez-paraswap, line 538: Is this intended change? It's so radical. Have you ran all tests to see all of them pass?

tsconfig.json
```diff
@@ -18,6 +18,6 @@
     "typeRoots": ["node_modules/@types"],
     "esModuleInterop": true
   },
-  "include": ["src/**/*"],
+  "include": ["src/**/*", "tests/**/*"],
```
erez-paraswap, line 21: now we include tests/** in build/dist? Why is it changed?

erez-paraswap (changes requested):

tsconfig.json
Verisana: I think I can remove it. It was done by @bogdoslavik because he has issues with IDE. After this we have:
```
"exclude": ["**/*.test.ts"]
```
So, it would add only test utils to build, though we don't need them.
Fixed

tests/constants-e2e.ts
Verisana: So, the reason to change was because some test were failing :)

tsconfig.json
```diff
@@ -5,7 +5,7 @@
     "module": "commonjs",
     "noImplicitAny": true,
     "outDir": "build",
-    "rootDir": "src",
+    "rootDirs": ["src"],
```
ColonelJ, line 8: Why this change?  `rootDir` and `rootDirs` are different things, and I'm not sure how it affects the build output...

src/dex/balancer-v2/balancer-v2.ts
```diff
@@ -18,26 +18,25 @@ import {
   MAX_INT,
   MAX_UINT,
   Network,
+  subgraphTimeout,
```
ColonelJ, line 21: I think this was renamed in `constants.ts`, but not here and other files.

Verisana: Fixed

tsconfig.json
Verisana: Fixed to previous state

src/dex/uniswap-v2/constants.ts
```diff
@@ -15,6 +16,7 @@ export const AllUniswapForks = [
   ...Object.keys(BiSwapConfig),
   ...Object.keys(DfynConfig),
   ...Object.keys(ExcaliburConfig),
+  ...Object.keys(DystopiaConfig),
```
ColonelJ, line 19: I think Dystopia should not be included in `AllUniswapForks` nor `UniswapForksWithNetwork` because it is not a standard fork, and it will break things if it gets mixed with another fork as `UniswapForkOptimized`.  Also, it doesn't support buy.  Comment it out on both and include a note in the comment above which currently covers `BakerySwap`.

ColonelJ (changes requested):

src/dex/uniswap-v2/constants.ts
Verisana: Fixed

ColonelJ (approved):

ColonelJ (changes requested):
I see very wrong prices in staging, something is wrong with stable pool price calculations.

ColonelJ (approved):

PR #127, opened 2022-06-20
fix: fix ABI issue for Dystopia

Verisana:
I put `decoderABI` argument not as the last element in `UniswapV2`, because didn't want to override many default params. I checked, it shouldn't break anything.

### DIFF
--- a/src/dex/uniswap-v2/uniswap-v2.ts
+++ b/src/dex/uniswap-v2/uniswap-v2.ts
@@ -1,4 +1,4 @@
-import { AbiCoder, Interface, JsonFragment } from '@ethersproject/abi';
+import { AbiCoder, Interface } from '@ethersproject/abi';
 import _ from 'lodash';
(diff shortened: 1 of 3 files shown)


erez-paraswap (approved):

src/dex/uniswap-v2/uniswap-v2.ts
```diff
@@ -105,6 +107,8 @@ export class UniswapV2EventPool extends StatefulEventSubscriber<UniswapV2PoolSta
         ' pool',
       logger,
     );
+
+    this.iface = new Interface(decoderABI);
```
ColonelJ, line 111: This is actually very inefficient, you are constructing a new Interface for every pool.  You could pass in the `Interface` directly instead of `JsonFragment[]` and creating it every time in the constructor.

ColonelJ (changes requested):

ColonelJ (approved):

PR #128, opened 2022-06-21
Add support for UniV3 multihop 

shresthagrawal:


Verisana (approved):
Looks good to me

src/dex/uniswap-v3.ts
```diff
@@ -67,6 +64,28 @@ export class UniswapV3
...
+    for (let i = 0; i < path.length; i++) {
+      const { tokenIn, tokenOut, fee } = path[i];
+      const [a, b] =
+        side === SwapSide.SELL ? [tokenIn, tokenOut] : [tokenOut, tokenIn];
+      if (i === 0) encodedPath = (encodedPath << 160n) | BigInt(a);
+      encodedPath = (encodedPath << 24n) | BigInt(fee);
+      encodedPath = (encodedPath << 160n) | BigInt(b);
+    }
+    const hexString = encodedPath.toString(16);
+    return (hexString.length % 2 ? '0x0' : '0x') + hexString;
```
ColonelJ, line 86: This code isn't good, if the first token starts with `0x00` (two or more zeroes) the `hexString` will be the wrong length for the data.  Should calculate the correct length (`43 * path.length * 2` I think) and pad with zeroes to that length.

ColonelJ (changes requested):

src/dex/uniswap-v3.ts
shresthagrawal: Very good finding! Will fix it

src/dex/uniswap-v3.ts
```diff
@@ -67,6 +64,29 @@ export class UniswapV3
...
+      tokenIn: Address;
+      tokenOut: Address;
+      fee: number;
+    }[],
+    side: SwapSide,
+  ): string {
+    let encodedPath = 0n;
+    const _path = side === SwapSide.SELL ? path : path.reverse();
+    for (let i = 0; i < path.length; i++) {
+      const { tokenIn, tokenOut, fee } = path[i];
```
Verisana, line 78: I think, here should be `_path[i]`

ColonelJ (approved):

PR #129, opened 2022-06-21
feat/584 - Add `Dex` for `Limit Order`

Verisana:
Resolves [BACK-584](https://linear.app/paraswap/issue/BACK-584/add-limit-order-dex-to-dexlib)
Related to [PR#520](https://github.com/paraswap/paraswap-api/pull/520)

### DIFF
--- a/src/dex/index.ts
+++ b/src/dex/index.ts
@@ -112,7 +112,7 @@ const AdapterNameAddressMap: {
   },
   [Network.AVALANCHE]: {
     AvalancheAdapter01: '0xb41Ec6e014e2AD12Ae8514216EAb2592b74F19e7',
-    AvalancheBuyAdapter: '0x05d0c2b58fF6c05bcc3e5F2D797bEB77e0A4CC7b',
+    AvalancheBuyAdapter: '0xe92b586627ccA7a83dC919cc7127196d70f55a06',
   },
   [Network.FANTOM]: {
     FantomAdapter01: '0xF52523B9d788F4E2Dd256dc5077879Af0448c37A',
--- a/src/dex/index.ts
+++ b/src/dex/index.ts
@@ -42,6 +42,7 @@ import { Platypus } from './platypus/platypus';
 import { GMX } from './gmx/gmx';
 import { WooFi } from './woo-fi/woo-fi';
 import { Dystopia } from './uniswap-v2/dystopia/dystopia';
+import { ParaswapLimitOrders } from './paraswap-limit-orders/paraswap-limit-orders';
 import { AugustusRFQOrder } from './augustus-rfq';
 
 const LegacyDexes = [
@@ -84,6 +85,7 @@ const Dexes = [
   GMX,
   WooFi,
   Dystopia,
+  ParaswapLimitOrders,
 ];
 
 const AdapterNameAddressMap: {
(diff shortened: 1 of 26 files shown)


src/dex/index.ts
```diff
@@ -105,8 +107,8 @@ const AdapterNameAddressMap: {
     BscBuyAdapter: '0xdA0DAFbbC95d96bAb164c847112e15c0299541f6',
   },
   [Network.ROPSTEN]: {
-    RopstenAdapter01: '0x74fF86C61CF66334dCfc999814DE4695B4BaE57b',
-    RopstenBuyAdapter: '0xDDbaC07C9ef96D6E792c25Ff934E7e111241BFf1',
+    RopstenAdapter01: '0x59b7F6258e78C3E5234bb651656EDd0e08868cd5',
+    RopstenBuyAdapter: '0x63e908A4C793a33e40254362ED1A5997a234D85C',
```
ColonelJ, line 111: Should update the adapters on all chains, they were deployed and whitelisted already.

src/dex/paraswap-limit-orders/paraswap-limit-orders.ts
```diff
@@ -0,0 +1,417 @@
...
+
+      const _srcAddress = _srcToken.address.toLowerCase();
+      const _destAddress = _destToken.address.toLowerCase();
+
+      if (_srcAddress === _destAddress) return null;
+
+      const expectedIdentifier = this.getIdentifier(_srcAddress, _destAddress);
+      if (
+        limitPools === undefined ||
+        !limitPools.some(p => p === expectedIdentifier)
```
ColonelJ, line 129: This is wrong logic, if `limitPools` is `undefined` then the pool can be used.  Check the logic in other DEX.

src/dex/paraswap-limit-orders/paraswap-limit-orders.ts
```diff
@@ -0,0 +1,417 @@
...
+
+  async getPricesVolume(
+    srcToken: Token,
+    destToken: Token,
+    amounts: bigint[],
+    side: SwapSide,
+    blockNumber: number,
+    limitPools?: string[],
+  ): Promise<null | ExchangePrices<ParaswapLimitOrdersData>> {
+    if (side === SwapSide.BUY) return null;
```
ColonelJ, line 115: We should be supporting `BUY` for this.

ColonelJ (changes requested):

src/dex/paraswap-limit-orders/paraswap-limit-orders.ts
Verisana: It was moved outside of current cycle as separate task after going prod:
https://linear.app/paraswap/issue/BACK-595/add-buy-suport-for-dex-limitorders

@erez-paraswap , is it ok or better to add to this cycle? It will involve changes on LO Service side too

src/dex/index.ts
Verisana: Updated adapters with addresses from here:
https://github.com/paraswap/paraswap-contracts/tree/master/deployments

And added new configs for all chains with `rfqAddress` from here:
https://github.com/paraswap/paraswap-limit-orders/tree/master/deployments

src/dex/paraswap-limit-orders/paraswap-limit-orders.ts
erez-paraswap: Let's leave it out of scope now and ship the sell side first. We'll address right after we ship and by then, LO will be out too

src/dex/paraswap-limit-orders/paraswap-limit-orders.ts
Verisana: Fixed

Louis-Amas (approved):

src/dex/paraswap-limit-orders/config.ts
```diff
@@ -0,0 +1,53 @@
...
+  [Network.FANTOM]: {
+    [SwapSide.SELL]: [{ name: 'FantomAdapter01', index: 6 }],
+    [SwapSide.BUY]: [{ name: 'FantomBuyAdapter', index: 2 }],
+  },
+  [Network.MAINNET]: {
+    [SwapSide.SELL]: [{ name: 'Adapter03', index: 9 }],
+    [SwapSide.BUY]: [{ name: 'BuyAdapter', index: 5 }],
+  },
+  [Network.POLYGON]: {
+    [SwapSide.SELL]: [{ name: 'PolygonAdapter01', index: 14 }],
```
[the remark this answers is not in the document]
Verisana: Good catch as always, fixed

PR #136, opened 2022-06-27
Feat: Synthetix atomic swaps

0xwontonsoup:
@Verisana this is currently a draft PR but i need help to understand _more thoroughly_ how to fully cover the integration for every Synthetix asset

I created a simple test as an example for what i think my integration test should look like. 
What am i missing so that I can do something similar with `DummyHexHelper`?

Thank you in advance :)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,23008 @@
+{
+  "name": "@paraswap/dex-lib",
+  "version": "2.0.57",
+  "lockfileVersion": 2,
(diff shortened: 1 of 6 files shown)


Verisana:
Hey @0xwontonsoup ,
Sorry for late reply. I missed your message and saw it only when we I was checking waiting PRs. I suppose the questions are irrelevant now, because we agreed that we will do the integration ourselves :)

PR #137, opened 2022-06-28
[Woofi-Polygon] add woofi contracts and token config for Polygon

fb-alexcq:
Woofi is releasing polygon deployment this week. Adding the woofi related smart contracts and token configs here.


Verisana:
@fb-alexcq , thanks for preparing PR 👍
I will try to merge as soon as our adapters will be updated on Polygon

PR #139, opened 2022-06-29
BACK-637 - Add WooFi config for Polygon

Verisana:
Changes from @fb-alexcq [PR#137](https://github.com/paraswap/paraswap-dex-lib/pull/137)

+ Fix for `getTopPools`: allow calling this function before `initializePricing`
+ Add tests for new network

### DIFF
--- a/src/dex/woo-fi/woo-fi.ts
+++ b/src/dex/woo-fi/woo-fi.ts
@@ -188,6 +188,10 @@ export class WooFi extends SimpleExchange implements IDex<WooFiData> {
     return Object.values(this.config.baseTokens);
   }
 
+  get isRefInfosEmpty(): boolean {
+    return Object.keys(this._refInfos).length === 0;
+  }
+
   protected _fillTokenInfoState(
     state: PoolState,
     address: string,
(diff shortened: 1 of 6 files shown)


src/dex/woo-fi/woo-fi.ts
```diff
@@ -565,6 +569,10 @@ export class WooFi extends SimpleExchange implements IDex<WooFiData> {
   }
 
   async updatePoolState(): Promise<void> {
+    if (this.isRefInfosEmpty) {
```
ColonelJ, line 572: This is a function, did you mean to call it here?

ColonelJ (changes requested):

src/dex/woo-fi/woo-fi.ts
Verisana: Yes, thank you for catching it 🥇

Verisana: @ColonelJ 
Do we need to call getter property?

```
  get isRefInfosEmpty(): boolean {
    return Object.keys(this._refInfos).length === 0;
  }
```

ColonelJ (approved):

erez-paraswap (approved):

PR #147, opened 2022-07-12
Feature/back 657

erez-paraswap:
Fix: Embr subgraph endpoint


Verisana (approved):

Louis-Amas (approved):

ColonelJ (approved):
Not sure we should be enabling this Embr because of low liquidity, but I'll approve anyway.

PR #154, opened 2022-07-27
SwaapV1 - Integration

borelien:
website: https://swaap.finance
app: https://beta.swaap.finance
docs: https://docs.swaap.finance
github: https://github.com/swaap-labs

### DIFF
--- a/src/dex/swaap-v1/swaap-v1-pool.ts
+++ b/src/dex/swaap-v1/swaap-v1-pool.ts
@@ -1,0 +1,897 @@
+import { DeepReadonly } from 'ts-essentials';
+import { BlockHeader } from '../../types';
+import { Address } from '../../types';
+import { IDexHelper } from '../../dex-helper/idex-helper';
+import {
+  SwaapV1PoolState,
+  SubgraphPoolBase,
+  SwaapV1PoolParameters,
+  SwaapV1PoolLiquidities,
+  SwaapV1PoolOracles,
+  OracleInitialData,
+  ChainLinkData,
+} from './types';
+import { StatefulEventSubscriber } from '../../stateful-event-subscriber';
+import { Log, MultiCallInput, MultiCallOutput } from '../../types';
+import PoolABI from '../../abi/swaap-v1/pool.json';
+import ProxyABI from '../../abi/chainlink.json';
+import Erc20ABI from '../../abi/erc20.json';
+import { Interface } from '@ethersproject/abi';
+
+export class SwaapV1Pool extends StatefulEventSubscriber<SwaapV1PoolState> {
+  static readonly poolInterface = new Interface(PoolABI);
+  static readonly proxyInterface = new Interface(ProxyABI);
+  static readonly erc20Interface = new Interface(Erc20ABI);
+  // LOG_CALL Event topics 0
+  static readonly SWAP_FEE_TOPIC =
+    '0x34e1990700000000000000000000000000000000000000000000000000000000';
+  static readonly LB_IN_ROUND_TOPIC =
+    '0x3cc3396000000000000000000000000000000000000000000000000000000000';
+  static readonly LB_STEP_IN_ROUND_TOPIC =
+    '0x045dbae000000000000000000000000000000000000000000000000000000000';
+  static readonly Z_TOPIC =
+    '0x3c515d9900000000000000000000000000000000000000000000000000000000';
+  static readonly HORIZON_TOPIC =
+    '0x314d204f00000000000000000000000000000000000000000000000000000000';
+  static readonly LB_IN_SEC_TOPIC =
+    '0x538e285c00000000000000000000000000000000000000000000000000000000';
+  static readonly MAX_UNPEG_TOPIC =
+    '0xacb21b3100000000000000000000000000000000000000000000000000000000';
+
+  static readonly ANSWER_UPDATED_TOPIC =
+    SwaapV1Pool.proxyInterface.getEventTopic('AnswerUpdated');
+  public addressSubscribers: Address[];
+  public oracleTokenMap: { [oracleAddress: Address]: Address };
+
+  constructor(
+    dexKey: string,
+    network: number,
+    protected pool: SubgraphPoolBase,
+    protected dexHelper: IDexHelper,
+  ) {
+    super(
+      `${dexKey}-${pool.id}`,
+      dexHelper.getLogger(`${dexKey}-${network}-${pool.id}`),
+    );
+
+    this.addressSubscribers = [pool.id];
+    this.oracleTokenMap = pool.tokens.reduce((acc, t) => {
+      this.addressSubscribers.push(
+        t.oracleInitialState.aggregator.toLowerCase(),
+      );
+      return {
+        ...acc,
+        [t.oracleInitialState.aggregator.toLowerCase()]: t.address,
+      };
+    }, {});
+  }
+
+  // Function which transforms the given state for the given log event.
+  // If the provided log does not affect the state, return null.
+  public async processLog(
+    state: DeepReadonly<SwaapV1PoolState>,
+    log: Readonly<Log>,
+    blockHeader: Readonly<BlockHeader>,
+  ): Promise<DeepReadonly<SwaapV1PoolState> | null> {
+    try {
+      const tokenAddress = this.oracleTokenMap[log.address.toLowerCase()];
+
+      if (tokenAddress) {
+        // log comes from a subscribed oracle
+
+        if (log.topics[0] != SwaapV1Pool.ANSWER_UPDATED_TOPIC) return null;
+        const decoded = SwaapV1Pool.proxyInterface.decodeEventLog(
+          'AnswerUpdated',
+          log.data,
+          log.topics,
+        );
+        const latestRoundId = Number(decoded.roundId.toString());
+        const latestRoundData = {
+          answer: BigInt(decoded.current.toString()),
+          timestamp: BigInt(decoded.updatedAt.toString()),
+        };
+
+        let previousRoundId = state.oracles[tokenAddress].latestRoundId;
+
+        if (previousRoundId == latestRoundId) {
+          return null;
+        }
+
+        let updatedHistoricalState = {
+          ...state.oracles[tokenAddress].historicalOracleState,
+        };
+        const roundLookbackWindow =
+          (state.parameters.priceStatisticsLookbackInRound - 1) *
+          state.parameters.priceStatisticsLookbackStepInRound;
+
+        // Case where roundIds are skipped
+        while (latestRoundId > previousRoundId + 1) {
+          delete updatedHistoricalState[previousRoundId - roundLookbackWindow];
+          updatedHistoricalState[++previousRoundId] = {
+            answer: 0n,
+            timestamp: 0n,
+          };
+        }
+
+        delete updatedHistoricalState[previousRoundId - roundLookbackWindow];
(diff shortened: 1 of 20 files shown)


src/dex/swaap-v1/swaap-v1.ts
```diff
@@ -0,0 +1,633 @@
...
+    amounts: bigint[],
+    side: SwapSide,
+    blockNumber: number,
+    _currentTimeStamp: bigint | null,
+    limitPools?: string[],
+  ): Promise<null | ExchangePrices<SwaapV1Data>> {
+    const from = this.dexHelper.config.wrapETH(_from);
+    const to = this.dexHelper.config.wrapETH(_to);
+
+    const topPools = await this.getTopPoolsForTokens(
```
Louis-Amas, line 562: `getTopPoolsForTokens` is doing http request to query the topPools. This is not good because the getPricesVolumes function is getting call a bunch of time on pricing request. You should be able to compute the topPools directly from memory

src/dex/swaap-v1/swaap-v1.ts
```diff
@@ -0,0 +1,633 @@
...
+    srcTokenAddress: Address,
+    destTokenAddress: Address,
+  ): Address[] {
+    if (!this.allPools) return [];
+    return Array.from(this.allPools.values())
+      .filter(
+        poolConfig =>
+          poolConfig.tokens.map(t => t.address).includes(srcTokenAddress) &&
+          poolConfig.tokens.map(t => t.address).includes(destTokenAddress),
+      )
```
Louis-Amas, line 152: This is not really efficient if we have lot of pools. Maybe it can be improve ?

Louis-Amas: Using another data structure than an array?

src/dex/swaap-v1/swaap-v1.ts
```diff
@@ -0,0 +1,633 @@
...
+    );
+  }
+
+  // // Returns list of top pools based on liquidity. Max
+  // // limit number pools should be returned.
+  async getTopPoolsForToken(
+    tokenAddress: Address,
+    count: number,
+  ): Promise<PoolLiquidity[]> {
+    return this.getTopPoolsForTokens([tokenAddress], count, null).then(
```
Louis-Amas, line 269: Here the usage of an http request is ok because we call getTopPoolsForToken only once in a while.

src/dex/swaap-v1/swaap-v1.ts
```diff
@@ -0,0 +1,633 @@
...
+        });
+      },
+      new Map<string, SubgraphPoolBase>(),
+    );
+  }
+
+  async getAllPools(
+    blockNumber: number,
+  ): Promise<Map<string, SubgraphPoolBase>> {
+    return await this.fetchAllSubgraphPools(blockNumber);
```
Louis-Amas, line 446: Remove this function and call it directly from init() one

src/dex/swaap-v1/swaap-v1.ts
```diff
@@ -0,0 +1,633 @@
...
+  // for pricing requests. It is optional for a DEX to
+  // implement this function
+  async initializePricing(blockNumber: number) {
+    await this.init(blockNumber);
+    if (!this.allPools)
+      throw new Error(
+        'initializePricing: SwaapV1 cfgInfo still null after init',
+      );
+    for (const poolConfig of this.allPools.values()) {
+      await this.addPool(poolConfig, blockNumber);
```
Louis-Amas, line 108: On the initialization it could be good to have a multicall to fetch the state of all pools. Otherwise on startup we bombard rpc node for no added value.

src/dex/swaap-v1/swaap-v1.ts
```diff
@@ -0,0 +1,633 @@
...
+  }
+
+  async addPool(poolConfig: SubgraphPoolBase, blockNumber: number) {
+    const pool = new SwaapV1Pool(
+      this.dexKey,
+      this.network,
+      poolConfig,
+      this.dexHelper,
+    );
+    const state = await pool.generateState(blockNumber);
```
Louis-Amas, line 119: I would add an optional parameter state to this function, and I will call generateState if it's not provided. This will permit you to batch the initial setState but fallback to generateState when a new pool is added

src/dex/swaap-v1/swaap-v1.ts
```diff
@@ -0,0 +1,633 @@
...
+
+    const srcTokenAddress = from.address.toLowerCase();
+    const destTokenAddress = to.address.toLowerCase();
+
+    if (srcTokenAddress === destTokenAddress) return null;
+    const currentTimestamp =
+      _currentTimeStamp || BigInt(Math.floor(Date.now() / 1000)); // the on-chain pricing logic is not too much sensitive to that value
+    return await Promise.all(
+      allowedPools.map(p =>
+        (async () => {
```
Louis-Amas, line 587: Why don't you use async directly on the (p = >) fn?

src/dex/swaap-v1/swaap-v1.ts
```diff
@@ -0,0 +1,633 @@
...
+          }
+          return {
+            prices,
+            unit,
+            data: {
+              pool: poolAddress,
+            },
+            poolAddresses: [poolAddress],
+            exchange: this.dexKey,
+            gasCost: 610 * 1000,
```
Louis-Amas, line 626: `gasCost: 610 * 1000` is it always the case ? Can you also create a global variable in capital case with this value?

`const SWAAPV1_GAS_COST = 610 * 1000;`

src/dex/swaap-v1/swaap-v1-pool.ts
```diff
@@ -0,0 +1,897 @@
...
+      );
+      return {
+        ...acc,
+        [t.oracleInitialState.aggregator.toLowerCase()]: t.address,
+      };
+    }, {});
+  }
+
+  // Function which transforms the given state for the given log event.
+  // If the provided log does not affect the state, return null.
```
Louis-Amas, line 70: If the provided log does not affect the state you should return previous state

ColonelJ: Returning `null` is actually correct if the state is not changed.  Doing so avoids unnecessarily creating an entry in the state history.

borelien: ok perfect!

src/dex/swaap-v1/swaap-v1.ts
borelien: Actually this number is probably too high.
But generally speaking our gas consumption might double depending on the state of the pool (let’s say from 250k to 500k).
And so we rather prefer to have a constant 500k gas instead of having failure because the state changed between the pricing and the execution (for instance during pricing the gas was 250k but during execution it’s now 500k and so reverts because on 'out-of-gas').
But maybe you have mechanisms to ensure this won’t happen, for instance you might add a margin on the gas already, and in that case we might no need this precaution?

src/dex/swaap-v1/swaap-v1.ts
borelien: true

src/dex/swaap-v1/swaap-v1.ts
borelien: KyberDMM impl of the findPools is quite similar (actually we copied their impl).
Is it fine if we add an intermediary variable like this?
```.filter(
        poolConfig => {
            const tokenAddresses = poolConfig.tokens.map(t => t.address)
            return tokenAddresses.includes(srcTokenAddress) && tokenAddresses.includes(destTokenAddress)
          }
        ,
      )
```

src/dex/swaap-v1/swaap-v1.ts
borelien: Is it a blocking point? We are not planning to support many pools in the future.

src/dex/swaap-v1/swaap-v1.ts
borelien: We had a discussion regarding with the Paraswap team on TG .
To give you the context:
1. we have a dynamic list of pools, meaning we can’t know in advance what would be the pools
2. unfortunately we don’t have a factory method to retrieve this list
3. so we rely on subgraph to get this list (and to add new pools to the subscriber)
4. what would be your recommandation?
5. at the time there was 2 options:
a) to rely on the getTopPoolsForToken call to update ouf pools:  but in that case we would need this function to be called frequently (let’s say on a daily or weekly basis), to be sure that new pools are listed quickly on Paraswap.
b) to update the list of pools directly in the getPricesVolume: we chose this one because we understood that it was fine and that time (but maybe we misunderstood).

Would it be doable to stick to the option b. but with a condition on blockNumber so that we don't request TheGraph more than once by block?

Louis-Amas: I think the solution **a** is the best solution, `getPricesVolume` is not made to handle this task. I think you can use a recursive `setTimeout` inside the `initializePricing` to do this task. 
Also new pool will probably have low liquidity so it will not impact the pricing that much.

src/dex/swaap-v1/swaap-v1.ts
Louis-Amas: How much pool do you have? We have multiple instance of api running dex-lib at the same time, and when we receive more pricing request we increase the amount of api instances to handle the load this will cause a big amount of rpc call congesting even more our rpc.

borelien: we have only 1 pool for now.

src/dex/swaap-v1/swaap-v1.ts
borelien: ok thanks!

src/dex/swaap-v1/swaap-v1.ts
borelien: ok will do it

src/dex/swaap-v1/swaap-v1.ts
borelien: not sure to understand what would be the use case of "This will permit you to batch the initial setState but fallback to generateState when a new pool is added"?

src/dex/swaap-v1/swaap-v1.ts
Louis-Amas: Ok then it's ok if you don't plan to have thousands pools :)

src/dex/swaap-v1/swaap-v1.ts
Louis-Amas: Related to https://github.com/paraswap/paraswap-dex-lib/pull/154#discussion_r1004688618
You can ignore it now :)

src/dex/swaap-v1/swaap-v1.ts
Louis-Amas: Are you able to predict the gas cost during a pricing ?
You can provide an array of gasCost representing the gas cost needed for every elements inside the amounts array. More precise you gas cost estimation are more likely our algorithm will choose your dex. We don't add any marging ourself but you can provide one if your prediction are not perfect :)

ColonelJ: Hi, the gas costs returned from pricing are just for a rough estimate, it is not used for transaction execution, where we do a proper estimateGas on the whole transaction and add a large margin of error.

If you are able to determine a gas cost from the state of the pool I'd encourage you to do that, otherwise you could just split the difference and use something like 375k.  You are not doing yourself any favours by setting the value too high, it will just make your DEX less likely to be chosen.

PR #66, opened 2022-03-18
feature/back-432 - migrate BalancerV1 to dex-lib interface

Verisana:

PR #157, opened 2022-07-28
Feature/back 432 followup - balancerV1

mwamedacen:
+ https://github.com/paraswap/paraswap-dex-lib/pull/66 synced to master 
+ fix tests 
+ slicing multicall rpc calls


src/dex/balancer-v1/config.ts
```diff
@@ -0,0 +1,57 @@
...
+  '0x34e1990700000000000000000000000000000000000000000000000000000000',
+  '0x49b5955200000000000000000000000000000000000000000000000000000000',
+  '0x4bb278f300000000000000000000000000000000000000000000000000000000',
+  '0x3fdddaa200000000000000000000000000000000000000000000000000000000',
+  '0xe4e1e53800000000000000000000000000000000000000000000000000000000',
+  '0xcf5e7bd300000000000000000000000000000000000000000000000000000000',
+  '0x02c9674800000000000000000000000000000000000000000000000000000000',
+];
+
+export const BALANCER_CHUNKS = 10;
```
Verisana, line 23: It is event base implementation. I think we can remove CHUNKING related logic

src/dex/balancer-v1/config.ts
```diff
@@ -0,0 +1,57 @@
...
+export const BALANCER_SWAP_GAS_COST = 120 * 1000;
+
+export const poolUrls: { [key: number]: string } = {
+  1: 'https://storageapi.fleek.co/balancer-bucket/balancer-exchange/pools',
+  42: 'https://storageapi.fleek.co/balancer-bucket/balancer-exchange-kovan/pools',
```
Verisana, line 28: Maybe better to use `Network.chainId`

src/dex/balancer-v1/config.ts
```diff
@@ -0,0 +1,57 @@
...
+};
+
+export const defaultfactoryAddress =
+  '0x9424B1412450D0f8Fc2255FAf6046b98213B76Bd';
+export const defaultMulticallAddress =
+  '0x514053acec7177e277b947b1ebb5c08ab4c4580e';
+
+export const POOL_FETCH_TIMEOUT = 5000;
+export const SUBGRAPH_TIMEOUT = 1000 * 10;
```
Verisana, line 37: There is a global `SUBGRAPH_TIMEOUT`. Please, use it

src/implementations/local-paraswap-sdk.ts
```diff
@@ -99,9 +99,15 @@ export class LocalParaswapSDK implements IParaSwapSDK {
     );
 
     if (!poolPrices || poolPrices.length == 0)
-      throw new Error('Fail to get price for ' + this.dexKey);
+      throw new Error('Fail to get prices for ' + this.dexKey);
+
+    const finalPrice = poolPrices?.find(
+      p => p.prices[p.prices.length - 1] !== BigInt(MAX_UINT),
```
Verisana, line 105: In some cases we can have `0n` as well. I think we should treat that case too. I have this change in other PR, but it wasn't merge to master yet:

```
    const finalPrice = poolPrices
      // Filter pools that didn't give latest output
      .filter(poolPrices => poolPrices.prices.slice(-1)[0] !== 0n)
      .sort((a, b) => {
        return side === SwapSide.SELL
          ? Number(b.prices.slice(-1)[0] - a.prices.slice(-1)[0])
          : Number(a.prices.slice(-1)[0] - b.prices.slice(-1)[0]);
      })[0];
```

I think we can unite and add case with BI_MAX_UINT. And you don't need to convert constant numbers to `bigint`. We have them in `bigint-constants.ts`

Verisana (commented):
I didn't go through balancer implementation itself. It is so big :(
But have a few things we can improve

src/implementations/local-paraswap-sdk.ts
mwamedacen: there you go 58be9e955d9ce9d9882702a53fcd3671f69d7c24

src/dex/balancer-v1/config.ts
mwamedacen: done b39a20849546ea7748e3fd8659e2545af1b05796

PR #166, opened 2022-08-15
BACK-677: Integrate Synthetix

Verisana:


src/dex/synthetix/config.ts
```diff
@@ -0,0 +1,55 @@
...
+    },
+    [Network.OPTIMISM]: {
+      readProxyAddressResolver: '0x1Cb059b7e74fD21665968C908806143E744D5F30',
+      flexibleStorage: '0x47649022380d182DA8010Ae5d257fea4227b21ff',
+      synths: [
+        { address: '0x8c6f28f2F1A3C87F0f938b96d27520d9751ec8d9', decimals: 18 }, // sUSD
+        { address: '0x00B8D5a5e1Ac97Cb4341c4Bc4367443c8776e8d9', decimals: 18 }, // sAAVE
+        { address: '0xB2b42B231C68cbb0b4bF2FFEbf57782Fd97D3dA4', decimals: 18 }, // sAVAX
+        { address: '0x298B9B95708152ff6968aafd889c6586e9169f1D', decimals: 18 }, // sBTC
+        { address: '0xE405de8F52ba7559f9df3C368500B6E6ae6Cee49', decimals: 18 }, // sBTC
```
ColonelJ, line 33: This is sETH

src/lib/decoders.ts
```diff
@@ -0,0 +1,118 @@
...
+export const uin256DecodeToBigNumber = (
+  result: MultiResult<string>,
+): BigNumber => {
+  if (!result.success) {
+    return BN_0;
+  }
+  return new BigNumber(
+    defaultAbiCoder.decode(['uint256'], result.returnData)[0],
+  );
+};
```
ColonelJ, line 47: This function doesn't work (you get NaN from `new BigNumber(ethers.utils.defaultAbiCoder.decode(['uint'], '0x0000000000000000000000000000000000000000000000000000000000000011')[0]).toString()`).  You can make it work by converting to string before BigNumber.

src/lib/decoders.ts
```diff
@@ -0,0 +1,118 @@
...
+  }
+  return parseFloat(defaultAbiCoder.decode(['uint256'], result.returnData)[0]);
+};
+
+export const uin128DecodeToFloat = (result: MultiResult<string>): number => {
+  if (!result.success) {
+    return 0;
+  }
+  return parseFloat(defaultAbiCoder.decode(['uint128'], result.returnData)[0]);
+};
```
ColonelJ, line 71: Don't understand why we are using `parseFloat` or having float methods at all (floats and ints are both just `number`).  The ABI coder will always return integers that will parse with `parseInt`.

src/lib/decoders.ts
```diff
@@ -0,0 +1,118 @@
...
+  return BigInt(defaultAbiCoder.decode(['uint'], result.returnData)[0]);
+};
+
+export const uint256ArrayDecode = (result: MultiResult<string>): bigint => {
+  if (!result.success) {
+    return 0n;
+  }
+  return defaultAbiCoder
+    .decode(['uint256[]'], result.returnData)[0]
+    .map((r: BigNumber) => BigInt(r.toString()));
```
ColonelJ, line 35: The `BigNumber` type in ethers is not `bignumber.js` but `BN.js`, which is completely different, so you have the wrong type here, but since we're just converting to string you could use `any` here.

ColonelJ (changes requested):
Some review on `decoders.ts`.

src/lib/decoders.ts
Verisana: I don't know. Better to check in the @Louis-Amas POC PR, where it is used. I just copied it here for consistency

src/lib/decoders.ts
Verisana: Changed to `any`

src/lib/decoders.ts
Verisana: Added `toString()`

src/dex/synthetix/config.ts
Verisana: Fixed

src/dex/synthetix/synthetix-state.ts
```diff
@@ -0,0 +1,933 @@
...
+    ).reduce<Record<string, boolean>>((acc, curr, i) => {
+      acc[curr] =
+        synthSuspensions[i] === true || synthExchangeSuspensions[i] === true;
+      return acc;
+    }, {});
+
+    const uniswapV3Slot0: Record<Address, Slot0> = {};
+    const tickCumulatives: Record<Address, Record<0 | 1, bigint>> = {};
+
+    _.chunk(slot0AndTickCumulatives, packCounter).map((result, i) => {
```
ColonelJ, line 207: Use `forEach` instead of `map` here?

Verisana: Your question lead me to google:
https://leanylabs.com/blog/js-forEach-map-reduce-vs-for-for_of/

And looks like `forEach` is almost two times faster than `map`. And `for of` is the fastest one. I didn't know about.
Fixed.

I always use `map` as rule of thumb, but shouldn't.

ColonelJ: `map` is for when you need an array of the output from running the function, if you don't need that, you can use `forEach`.  It's quite annoying that it's so much slower than `for ... of` and other plain loops though, maybe it's better to use those when you can...

src/lib/decoders.ts
Louis-Amas: I added this methods when migrating curve I think we were using parseFloat there but no idea why

Louis-Amas: Let's remove that if it's useless

src/dex/synthetix/synthetix-state.ts
```diff
@@ -0,0 +1,933 @@
...
+      );
+
+    const [block, observationsRoundDataAndOverridden] = await Promise.all([
+      this.dexHelper.web3Provider.eth.getBlock(blockNumber || 'latest'),
+      this.multiWrapper.tryAggregate<
+        OracleObservation | LatestRoundData | string
+      >(true, observationsRoundAndOverriddenCallData, blockNumber),
+    ]);
+
+    addressesFromPKBoundary = addressesFromPK.length * packCounter;
```
ColonelJ, line 230: I think this should be `_packCounter` not `packCounter` (turns out they are the same currently, so it happens to work).

src/dex/synthetix/synthetix.ts
```diff
@@ -0,0 +1,334 @@
...
+    protected dexKey: string,
+    readonly dexHelper: IDexHelper,
+    protected adapters = Adapters[network] || {},
+    readonly config = SynthetixConfig[dexKey][network],
+  ) {
+    super(dexHelper.config.data.augustusAddress, dexHelper.web3Provider);
+    this.config = this.normalizeConfig(this.config);
+    this.logger = dexHelper.getLogger(dexKey);
+    this.combinedIface = new Interface(CombinedSynthetixABI);
+    this.multiWrapper = new MultiWrapper(
```
Louis-Amas, line 58: We should probably instantiate one by chain

src/dex/synthetix/synthetix.ts
```diff
@@ -0,0 +1,334 @@
...
+      this.dexHelper.multiContract,
+      this.logger,
+    );
+    this.synthetixState = new SynthetixState(
+      this.dexKey,
+      this.dexHelper,
```
Louis-Amas, line 64: Probably in dex-helper

Louis-Amas (approved):

src/dex/synthetix/synthetix-state.ts
```diff
@@ -0,0 +1,933 @@
...
+        acc[Object.keys(aggregatorAddressesWithoutZeros)[i]] = {
+          latestRoundData: cur,
+          getRoundData,
+        };
+        return acc;
+      },
+      {},
+    );
+
+    const uniswapV3Observations = _.chunk(observations, packCounter).reduce<
```
ColonelJ, line 290: Again, should be `_packCounter` not `packCounter` I believe.

ColonelJ (changes requested):

src/dex/synthetix/synthetix.ts
Verisana: Yeah, you are right. Fixed

src/dex/synthetix/synthetix.ts
Verisana: Done

dex-template/__DexName__-e2e.test.ts.template
```diff
@@ -50,83 +50,110 @@ import { generateConfig } from '../../config';
   (This comment should be removed from the final implementation)
 */
 
+async function testForNetwork(
```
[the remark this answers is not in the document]
Verisana: Fixed

dex-template/__DexName__-integration.test.ts.template
```diff
@@ -87,116 +76,171 @@ async function checkOnChainPricing(
...
+  funcNameToCheck: string,
+) {
+  const networkTokens = Tokens[network];
 
-  beforeAll(async () => {
-    blockNumber = await dexHelper.web3Provider.eth.getBlockNumber();
+  const pools = await __DexNameCamel__.getPoolIdentifiers(
+    networkTokens[srcTokenSymbol],
+    networkTokens[destTokenSymbol],
+    SwapSide.SELL,
```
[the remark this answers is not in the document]
Verisana: Fixed

src/dex/synthetix/synthetix-state.ts
Verisana: Good catch, thanks! Fixed it 👍

src/dex/synthetix/synthetix-state.ts
Verisana: Good catch, thanks! Fixed it 👍

ColonelJ (approved):

Louis-Amas (approved):

PR #167, opened 2022-08-15
Feat/solidly-forks-event-based

mwamedacen:
resolves BACK-681


src/dex/index.ts
```diff
@@ -41,13 +40,12 @@ import { KyberDmm } from './kyberdmm/kyberdmm';
...
-import { Dystopia } from './uniswap-v2/dystopia/dystopia';
-import { SpiritSwapV2 } from './spirit-swap-v2';
 import { ParaSwapLimitOrders } from './paraswap-limit-orders/paraswap-limit-orders';
 import { AugustusRFQOrder } from './augustus-rfq';
 import Web3 from 'web3';
-import { Velodrome } from './velodrome';
-import { Solidly } from './solidly';
+import { Solidly } from './solidly/solidly';
+import { Velodrome } from './solidly/forks-override/velodrome';
+import { SpiritSwapV2 } from './solidly/forks-override/spiritSwapV2';
```
Verisana, line 48: I think you just need to import only main implementation, for example, Solidy (or Dystopia). Other forks should be picked up and added from config automatically from function:

```
  public static dexKeysWithNetwork: { key: string; networks: Network[] }[] =
    getDexKeysWithNetwork(SynthetixConfig);

```

src/dex/solidly/config.ts
```diff
@@ -0,0 +1,75 @@
...
+      // updatable fees on the factory without event
+      stableFee: 2,
+      volatileFee: 2,
+      poolGasCost: 180 * 1000,
+      feeCode: 2,
+    },
+  },
+};
+
+// TODO: segment adapters by dex key
```
Verisana, line 64: Why? It is ok, if they are just forks and we use one index for each of them

src/dex/solidly/forks-override/spiritSwapV2.ts
```diff
@@ -0,0 +1,60 @@
+import { Solidly } from '../solidly';
```
Verisana, line 1: This file may be a mistake. Is it possible to incorporate fee changes into one instance of `Solidly` which is initialized with different parameters for different versions?

Because we have inheritance: `SpiritSwapV2 -> Solidly -> UniswapV2 -> SimpleExchange`. It is becoming very long :(

Or maybe @ColonelJ  has other opinion

src/dex/solidly/forks-override/spiritSwapV2.ts
```diff
@@ -0,0 +1,60 @@
...
+const SpiritSwapV2PairABI = [
+  {
+    inputs: [],
+    name: 'fee',
+    outputs: [{ internalType: 'uint256', name: '', type: 'uint256' }],
+    stateMutability: 'view',
+    type: 'function',
+  },
+];
+
```
Verisana, line 19: ABI may be passed on initialization part. And you can pass `variation: 'solidly' | 'spiritswapv2' | 'velodrome'` to make a difference in fee calculation. I think it may be more clean

Verisana: You can have that `variation` variable in config

src/dex/solidly/solidly-e2e.test.ts
```diff
@@ -0,0 +1,613 @@
...
+
+    describe('Dystopia', () => {
+      const dexKey = 'Dystopia';
+      const usdAmount = '1000000';
+
+      describe('Dystopia UniswapV2 Pools', () => {
+        const maticAmount = '1000000000000000000';
+
+        describe('simpleSwap', () => {
+          it('MATIC -> TOKEN', async () => {
```
Verisana, line 331: I think better to have specific names instead of generic `TOKEN` because it is easy to run particular tests

src/dex/solidly/solidly-e2e.test.ts
```diff
@@ -0,0 +1,613 @@
...
+    const provider = new StaticJsonRpcProvider(
+      generateConfig(network).privateHttpProvider,
+      network,
+    );
+
+    describe('Velodrome', () => {
+      const dexKey = 'Velodrome';
+
+      describe('simpleSwap', () => {
+        it('NATIVE -> TOKEN', async () => {
```
Verisana, line 528: I think better to have specific names instead of generic TOKEN because it is easy to run particular tests

src/dex/solidly/solidly.ts
```diff
@@ -335,10 +345,14 @@ export class Dystopia extends UniswapV2 {
     count: number,
   ): Promise<PoolLiquidity[]> {
     if (!this.subgraphURL) return [];
+
+    const stableFieldKey =
+      this.dexKey.toLowerCase() === 'solidly' ? 'stable' : 'isStable';
+
     const query = `query ($token: Bytes!, $count: Int) {
```
Verisana, line 352: For `SpiritSwapV2` there is a difference: `$token: String`, not `Bytes!`. Better to check with tests other forks if there subgraph query compatible or not

Verisana (commented):
I don't see caching. I asked Louis about the changes when we have process log separation. Maybe we will not need it anymore and it will be done behind the scenes :)

src/dex/solidly/config.ts
mwamedacen: I was imaging a case we'd have to do it but probably not relevant

src/dex/index.ts
mwamedacen: Nop it's purposeful, mind:
- omit here  https://github.com/paraswap/paraswap-dex-lib/blob/e971fb70bfe8dd4e94f23ad7c5975df6921a9a0e/src/dex/solidly/solidly.ts#L34
- pick here https://github.com/paraswap/paraswap-dex-lib/blob/e971fb70bfe8dd4e94f23ad7c5975df6921a9a0e/src/dex/solidly/forks-override/spiritSwapV2.ts#L24

Though echoes your other comment about how to implement propre override.

src/dex/solidly/forks-override/spiritSwapV2.ts
mwamedacen: I kinda mimicked what we did for uniswap-v2 forks that supports dynamic fees https://github.com/paraswap/paraswap-dex-lib/blob/feat/solidly-forks/src/dex/uniswap-v2/excalibur.ts#L45. I honestly prefer this design as it ends up with less spaghetti code

src/dex/solidly/solidly.ts
mwamedacen: That's unfortunate. Thx I will  check

src/dex/solidly/forks-override/spiritSwapV2.ts
mwamedacen: Though thinking of your variant idea. I can be generic implementations like SoldilyWithFeesOnFactory and SoldilyWithFeesOnPair

src/dex/solidly/solidly.ts
mwamedacen: as discussed IRL (reporting for posterity):
1/ it looks like `Bytes` and 'String' are interchangeable when subgraph is expecting a string.
2/ [this](https://github.com/paraswap/paraswap-dex-lib/pull/167/commits/08f6f610e58e7ddf4bbcb14fbec6b28e7e27893c) was missing

ColonelJ (approved):

Verisana (approved):

PR #171, opened 2022-08-19
Feature - Jarvis v6 Polygon Integration

FabienCoutant:
Documentations: [link](https://learn.jarvis.network/protocol-overview/minting )
Contract addresses: [link](https://gitlab.com/jarvis-network/apps/exchange/mono-repo/-/blob/fix/lendingManager/libs/contracts/networks/137.json )

**NOTE: when merged, we must disable the Jarvis V4 integration on polygon.** 

Do not hesitate if you need additional information.


src/dex/jarvis-v6/jarvis-v6.ts
```diff
@@ -0,0 +1,463 @@
...
+    const poolIdentifier = eventPool.getIdentifier();
+    if (limitPools && !limitPools.includes(poolIdentifier)) return null;
+
+    const poolState = await this.getPoolState(eventPool, blockNumber);
+
+    const unitVolume = getBigIntPow(
+      (side === SwapSide.SELL ? srcToken : destToken).decimals,
+    );
+
+    const swapFunction = getJarvisSwapFunction(srcToken, eventPool.poolConfig);
```
mwamedacen, line 167: imo, this should return `MINT | REDEEM | null` (third case if destToken doesn't match collateralToken) and we should early return to skip surrounding async calls (better move up).

src/dex/jarvis-v6/config.ts
```diff
@@ -0,0 +1,68 @@
+import { DexParams } from './types';
+import { DexConfigMap, AdapterMappings } from '../../types';
+import { Network, SwapSide } from '../../constants';
+import { Interface } from '@ethersproject/abi';
+import ChainlinkAccessControlledOffchainAggregatorABI from '../../abi/jarvis/ChainlinkAccessControlledOffchainAggregator.json';
+import SynthereumPriceFeedABI from '../../abi/jarvis/SynthereumPriceFeed.json';
+
+export const JarvisV6Config: DexConfigMap<DexParams> = {
+  JarvisV6: {
```
mwamedacen, line 9: only jEUR and jCHF will be swappable from/to ?

src/dex/jarvis-v6/utils.ts
```diff
@@ -0,0 +1,130 @@
...
+  );
+
+  const pricesFeedValues = await priceFeedContract.methods
+    .getLatestPrices(pairList)
+    .call({}, blockNumber);
+
+  return pricesFeedValues;
+}
+
+export function getJarvisPoolFromTokens(
```
mwamedacen, line 79: any way we can discover pools dynamically through onchain registry or similar?

src/dex/jarvis-v6/utils.ts
```diff
@@ -0,0 +1,130 @@
...
+}
+
+export function getJarvisSwapFunction(
+  srcToken: Token,
+  pool: PoolConfig,
+): JarvisSwapFunctions {
+  const srcAddress = srcToken.address.toLowerCase();
+  if (srcAddress === pool.collateralToken.address.toLowerCase())
+    return JarvisSwapFunctions.MINT;
+  return JarvisSwapFunctions.REDEEM;
```
mwamedacen, line 129: safer to check if `(srcToken|destToken)` is a `(jFiat|collateralToken)` (see my comment up in call stack)

src/dex/jarvis-v6/types.ts
```diff
@@ -0,0 +1,62 @@
...
+type JarvisV6RedeemParam = [
+  numTokens: string,
+  minCollateral: string,
+  expiration: string,
+  recipient: string,
+];
+
+export type JarvisV6Params = JarvisV6MintParam | JarvisV6RedeemParam;
+
+export enum JarvisSwapFunctions {
```
mwamedacen, line 59: I can't find equivalent method of `exchange()`, new protocol doesn't allow to exchange jFiat <> jFiat swaps ?

src/dex/jarvis-v6/jarvis-v6.ts
```diff
@@ -0,0 +1,463 @@
...
+      destToken,
+      destAmount,
+      swapData,
+      data.poolAddress.toLowerCase(),
+    );
+  }
+
+  // Returns list of top pools based on liquidity. Max
+  // limit number pools should be returned.
+  async getTopPoolsForToken(
```
mwamedacen, line 294: This should be implemented to help our multipath engine use jFiats as intermediary tokens. 
Also, as the protocol doesn't have `exchange()` anymore, I believe this will allow atomic swaps between jFiats.

src/dex/jarvis-v6/utils.ts
```diff
@@ -0,0 +1,130 @@
...
+      ? desireDecimal - currentDecimal
+      : currentDecimal - desireDecimal,
+  );
+
+  return isDecimalIncrease ? amount * bigIntPow : amount / bigIntPow;
+}
+
+// This is needed in order to not modify existing logic and use this wrapper
+// to be safe if we receive not cached decimals
+export function getBigIntPow(decimals: number): bigint {
```
mwamedacen, line 116: please reuse [this one](https://github.com/paraswap/paraswap-dex-lib/blob/6ced464284f7dadacb0a87d5a9630aff625a3afe/src/utils.ts#L53) (I must admit we need to clean our repo too ^^')

src/dex/jarvis-v6/config.ts
FabienCoutant: For now only jEUR and jCHF have been deployed with v6.

src/dex/jarvis-v6/utils.ts
FabienCoutant: Agree to move on a dynamic way to get pools info, but for the next version of the integration

src/dex/jarvis-v6/utils.ts
FabienCoutant: This function is only called when the pool has been already defined. I then supposed that both `srcAddress` and `destAddress` target a valid pool. Only the `MINT` or `REDEEM` info is required in order to know with function to call from the contract.

src/dex/jarvis-v6/types.ts
FabienCoutant: The `exchange()` was just the `mint() + redeem()` behind. We removed the `exhange()` method and will deploy another contract that will handle the `exchange()` things.

src/dex/jarvis-v6/jarvis-v6.ts
FabienCoutant: Cf. comment above:  another contract will be deployed to handle the swaps between jFiats

src/dex/jarvis-v6/utils.ts
mwamedacen: [this](https://github.com/paraswap/paraswap-dex-lib/pull/171#discussion_r964535355) works for me, ignore here

src/dex/jarvis-v6/types.ts
mwamedacen: Got you.
Sorry I meant "new contract" instead of new protocol.

src/dex/jarvis-v6/config.ts
FabienCoutant: Missing pools config have been added [here](https://github.com/paraswap/paraswap-dex-lib/pull/171/commits/80b30b2e89e0335a3d1a3533751f787053017bb2)

src/dex/jarvis-v6/utils.ts
```diff
@@ -0,0 +1,123 @@
...
+  const multiContract = dexHelper.multiContract;
+  const PoolCallData = poolConfigs
+    .map(pool => [
+      {
+        target: pool.address,
+        callData: poolInterface.encodeFunctionData('feePercentage', []),
+      },
+    ])
+    .flat();
+  const poolDataCalled = await multiContract.methods
```
mwamedacen, line 36: As discussed in dm, we can't query the aggregated priceFeed contract through our multicall contract due to whitelisting so 2 rpc are necessary  ✅ 

I suggest a minor micro-optimisation still:
```js
const [pricesFeedResults, poolDataCalled] = await Promise.all([
  _getPricesFeedValue(
    dexHelper, 
    poolConfigs, 
    priceFeedAddress, 
    blockNumber),
  multiContract
  .methods()
  .aggregate(PoolCallData)
  .call({}, blockNumber)
]);
```

src/dex/jarvis-v6/utils.ts
```diff
@@ -0,0 +1,123 @@
...
+    .aggregate(PoolCallData)
+    .call({}, blockNumber);
+
+  let i = 0;
```
mwamedacen, line 40: we can shoot this and rely on index

src/dex/jarvis-v6/config.ts
```diff
@@ -0,0 +1,308 @@
...
+            decimals: 18,
+            symbol: 'jEUR',
+          },
+          collateralToken: {
+            address: '0x2791bca1f2de4661ed88a30c99a7a9449aa84174',
+            decimals: 6,
+            symbol: 'USDC',
+          },
+          chainLink: {
+            interface: new Interface(
```
mwamedacen, line 30: it'll be less noisy to maintain only addresses here and instantiate  chainLink once only like [here](https://github.com/paraswap/paraswap-dex-lib/blob/8adeb6af1af93e7be7dfc8ba53d8f3f06cdadb50/src/dex/jarvis-v6/jarvis-v6.ts#L50).
Minor though

src/dex/jarvis-v6/utils.ts
mwamedacen: For posterity we can achieve this with:
- https://gitlab.com/jarvis-network/apps/exchange/mono-repo/-/blob/dev/libs/contracts/contracts/core/registries/Registry.sol
- https://gitlab.com/jarvis-network/apps/exchange/mono-repo/-/blob/dev/libs/contracts/contracts/core/registries/PoolRegistry.sol

as suggested by @Verisana

PR #172, opened 2022-08-24
Rollup gas estimation

ColonelJ:
Resolves BACK-701


mwamedacen (approved):

Louis-Amas (approved):

Louis-Amas (approved):

dex-template/__DexName__.ts.template
```diff
@@ -92,6 +94,14 @@ export class __DexName__
     return null;
   }
 
+  // Returns estimated gas cost of calldata for this DEX in multiSwap
```
Verisana, line 97: Maybe need to point in comment that it makes sense to change only if `Dex` supports `Optimism`?

Verisana (approved):

dex-template/__DexName__.ts.template
ColonelJ: I think we can have it correctly for all new DEX (not just Optimism or Arbitrum ones), since I already did them all the existing ones.

PR #170, opened 2022-08-19
Cone.exchange

bogdoslavik:
# Cone Exchange
https://www.cone.exchange/
### BSC
Solidly (actually Dystopia) fork but with variable fees.
Fee number used as divider.

Fees defaults:
- Stable: 10000 (0,01%) ('1' in uniswap)
- Volatile: 2000 (0,05%) ('5' in uniswap)

But fees may be changed.

### ParaSwapTODO: 
SimpleSwap tests passed, but for Muti and Mega swaps
!!! Cone needs same router like Dystopia, but on BSC network.

PR #176, opened 2022-08-27
Cone exchange 

bogdoslavik:
This is PR, replacing https://github.com/paraswap/paraswap-dex-lib/pull/170 

# Cone Exchange
https://www.cone.exchange/
### BSC
Solidly (actually Dystopia) fork but with variable fees.
Fee number used as divider.

Fees defaults:
- Stable: 10000 (0,01%) ('1' in uniswap)
- Volatile: 2000 (0,05%) ('5' in uniswap)

But fees may be changed.

### ParaSwapTODO: 
SimpleSwap tests passed, but for Muti and Mega swaps
!!! Cone needs same router like Dystopia, but on BSC network.

See `encodePools` at `cone.ts`

### DIFF
--- a/src/abi/uniswap-v2/ConeFactory.json
+++ b/src/abi/uniswap-v2/ConeFactory.json
@@ -1,0 +1,279 @@
+[
+  {
+    "inputs": [],
+    "stateMutability": "nonpayable",
+    "type": "constructor"
+  },
+  {
+    "anonymous": false,
+    "inputs": [
+      {
+        "indexed": true,
+        "internalType": "address",
+        "name": "token0",
+        "type": "address"
+      },
+      {
+        "indexed": true,
+        "internalType": "address",
+        "name": "token1",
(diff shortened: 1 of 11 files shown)


mwamedacen:
Everything is already here to integrate a solidly fork please refer to [this config file](https://github.com/paraswap/paraswap-dex-lib/blob/master/src/dex/solidly/config.ts). From my observations, the only discrepancy is that fees are handled on the pool. Refer to the SpiritSwapV2 integration [here](https://github.com/paraswap/paraswap-dex-lib/blob/master/src/dex/solidly/forks-override/spiritSwapV2.ts) (in your case you can return swapFee directly, no need to update feeFactory).

bogdoslavik:
Thanks for the helpful pointers to the right files. Development becomes so much easier with them :).

I created a new PR:  https://github.com/paraswap/paraswap-dex-lib/pull/177

PR #184, opened 2022-09-07
Feature/jarvis v6

mwamedacen:
orig PR: https://github.com/paraswap/paraswap-dex-lib/pull/171

closes BACK-548


src/dex/jarvis-v6/utils.ts
```diff
@@ -0,0 +1,123 @@
...
+  blockNumber: number | 'latest',
+) {
+  const pairList = poolConfigs.map(pool =>
+    ethers.utils.formatBytes32String(pool.priceFeedPair),
+  );
+
+  const priceFeedContract = new dexHelper.web3Provider.eth.Contract(
+    SynthereumPriceFeedABI as any,
+    priceFeedAddress,
+  );
```
ColonelJ, line 71: This is creating a new contract instance every time through this function.

src/dex/jarvis-v6/jarvis-v6.ts
```diff
@@ -0,0 +1,473 @@
...
+    side: SwapSide,
+    blockNumber: number,
+    limitPools?: string[],
+  ): Promise<null | ExchangePrices<JarvisV6Data>> {
+    const eventPool = this.getEventPool(srcToken, destToken);
+
+    if (!eventPool) return null;
+    const poolAddress = eventPool.poolConfig.address.toLowerCase();
+    const poolIdentifier = eventPool.getIdentifier();
+    if (limitPools && !limitPools.includes(poolIdentifier)) return null;
```
ColonelJ, line 160: This line should be commented out with a note that since this DEX is constant price and we have fixed routing (via USDC) which will cause problems for route aggregation (multiple routes with jToken -> USDC), we should not use pool allocations.

src/dex/jarvis-v6/jarvis-v6.ts
```diff
@@ -0,0 +1,473 @@
...
+  // any pools can be used.
+  // Pair
+  async getPricesVolume(
+    srcToken: Token,
+    destToken: Token,
+    amounts: bigint[],
+    side: SwapSide,
+    blockNumber: number,
+    limitPools?: string[],
+  ): Promise<null | ExchangePrices<JarvisV6Data>> {
```
ColonelJ, line 154: The BUY side (exact output) is not implemented by the protocol and should be disabled here to avoid issues.

src/dex/jarvis-v6/jarvis-v6.ts
```diff
@@ -0,0 +1,473 @@
...
+    side: SwapSide,
+  ): Promise<SimpleExchangeParam> {
+    const { swapFunction } = data;
+    const timestamp = (Date.now() / 1000 + THIRTY_MINUTES).toFixed(0);
+
+    let swapFunctionParams: JarvisV6Params;
+    switch (swapFunction) {
+      case JarvisSwapFunctions.MINT:
+        swapFunctionParams = [
+          destAmount,
```
ColonelJ, line 260: This is a minimum amount, can be set to 1 like in the contract.

src/dex/jarvis-v6/jarvis-v6.ts
```diff
@@ -0,0 +1,473 @@
...
+          srcAmount,
+          timestamp,
+          this.augustusAddress,
+        ];
+        break;
+      case JarvisSwapFunctions.REDEEM:
+        swapFunctionParams = [
+          srcAmount,
+          destAmount,
```
ColonelJ, line 269: As above, this is a minimum amount, and can be set to 1 to match the contract adapter.

src/dex/jarvis-v6/jarvis-v6.ts
```diff
@@ -0,0 +1,473 @@
...
+    }
+
+    this.logger.info(
+      `Get ${this.dexKey}_${this.network} MaxTokensCapacity from pool : ${poolAddress}`,
+    );
+
+    const poolContractInstance = new this.dexHelper.web3Provider.eth.Contract(
+      JarvisV6PoolABI as any,
+      poolAddress,
+    );
```
ColonelJ, line 325: May make sense to avoid creating contract instances here as well.

src/dex/jarvis-v6/jarvis-v6.ts
```diff
@@ -0,0 +1,473 @@
...
+  async getMaxTokensCapacity(
+    poolAddress: Address,
+    blockNumber: number,
+  ): Promise<bigint> {
+    const cacheKey = `${this.dexKey}_maxTokensCapacity_${poolAddress}`;
+    const cachedMaxTokensCapacity = await this.dexHelper.cache.get(
+      this.dexKey,
+      this.network,
+      cacheKey,
+    );
```
ColonelJ, line 309: I think the value should be cached in local memory as well for the same TTL at the time of request, otherwise when the remote cache expires, all instances may end up trying to repopulate it at the same time.

src/dex/jarvis-v6/jarvis-v6.ts
```diff
@@ -0,0 +1,473 @@
...
+    if (collateralDecimals === 18) return result;
+    return convertToNewDecimals(result, 18, collateralDecimals);
+  }
+
+  getCalldataGasCost(poolPrices: PoolPrices<JarvisV6Data>): number | number[] {
+    return (
+      CALLDATA_GAS_COST.DEX_OVERHEAD +
+      CALLDATA_GAS_COST.LENGTH_SMALL +
+      // opType
+      CALLDATA_GAS_COST.OFFSET_SMALL +
```
ColonelJ, line 468: `INDEX` would make more sense than `OFFSET_SMALL` since this is not an offset.

ColonelJ (changes requested):

src/dex/jarvis-v6/jarvis-v6.ts
mwamedacen: I feel like this should be handled at upper level and abstracted away on calling `cache.(get|.setex)`

src/dex/jarvis-v6/jarvis-v6.ts
mwamedacen: done buy this was not enough, had to keep my change on api side

src/dex-helper/icache.ts
```diff
@@ -12,4 +12,19 @@ export interface ICache {
     seconds: number,
     value: string,
   ): Promise<void>;
+
+  getAndCacheLocally(
+    dexKey: string,
+    network: number,
+    cacheKey: string,
+    ttl: number,
```
ColonelJ, line 20: I'd call it `seconds` for consistency with `setex`.

ColonelJ: Or change the others to `ttl` that works too...

mwamedacen: I thought I renamed ttl everywhere ^^'

Louis-Amas (approved):

src/dex/jarvis-v6/jarvis-v6.ts
ColonelJ: This doesn't appear changed.  I think it makes sense to save the instance for later use.

ColonelJ (changes requested):

src/dex/jarvis-v6/types.ts
```diff
@@ -0,0 +1,59 @@
...
+];
+
+export type JarvisV6Params = JarvisV6MintParam | JarvisV6RedeemParam;
+
+export enum JarvisSwapFunctions {
+  MINT = 'mint',
+  REDEEM = 'redeem',
+}
+
+export type JarvisV6SytemMaxVars = {
```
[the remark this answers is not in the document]
mwamedacen: fixed

ColonelJ (approved):

PR #194, opened 2022-09-20
Feat/back 727

Louis-Amas:


src/dex/balancer-v1/balancer-v1.ts
```diff
@@ -0,0 +1,919 @@
...
+    const tokenOut = event.args.tokenOut.toLowerCase();
+    const tokenAmountOut = new BigNumber(event.args.tokenAmountOut.toString());
+    _state.tokens = _state.tokens.map(token => {
+      if (token.address.toLowerCase() === tokenIn)
+        token.balance = new BigNumber(token.balance)
+          .plus(tokenAmountIn)
+          .toString();
+      else if (token.address.toLowerCase() === tokenOut)
+        token.balance = new BigNumber(token.balance)
+          .plus(tokenAmountOut)
```
ColonelJ, line 220: Should be minus...

ColonelJ (changes requested):

src/dex/balancer-v1/balancer-v1.ts
```diff
@@ -0,0 +1,919 @@
...
+  }
+
+  handleSwap(
+    event: LogDescription,
+    state: DeepReadonly<MinimalPoolState>,
+  ): MinimalPoolState {
+    const _state = _.cloneDeep(state) as MinimalPoolState;
+
+    const tokenIn = event.args.tokenIn.toLowerCase();
+    const tokenAmountIn = event.args.tokenAmountIn;
```
ColonelJ, line 209: Should be converted to bignumber.js BigNumber as `tokenAmountOut` is.

src/dex/balancer-v1/balancer-v1.ts
```diff
@@ -377,15 +377,20 @@ export class BalancerV1EventPool {
...
     amount: bigint,
   ) {
     if (
       side === SwapSide.BUY &&
-      amount * 2n > BigInt(poolData.balanceOut.toFixed(0))
+      amount * 2n >
+        BigInt(poolData.balanceOut.toString().slice(0, -to.decimals))
     ) {
+      console.log(poolData.balanceOut.toString(), amount * 2n);
+      console.log(BI_MAX_INT);
```
ColonelJ, line 393: I don't understand these changes.  `amount` is already in the token decimals so why are you removing digits?  BigNumber from bignumber.js `toString()` is not good to use because it gives exponential notation if the number is big, use `toFixed(0)` whenever you're converting integers from that library.  Console logs shouldn't be here, and probably in this condition returning `0n` is preferred to indicate no price.

src/dex/balancer-v1/balancer-v1-e2e.test.ts
```diff
@@ -134,7 +134,7 @@ describe('BalancerV1 E2E Mainnet', () => {
         tokens.USDC,
         tokens.ETH,
         holders.USDC,
-        '200000000000000000',
+        '200000000000',
```
ColonelJ, line 137: This value is in ETH since it's BUY, and the original amount was fine (0.2ETH).

ColonelJ (changes requested):

PR #199, opened 2022-09-26
BACK-738: Fix Curve pricing bug

Verisana:


src/dex/curve-v1/pools/3pool.ts
```diff
@@ -67,11 +67,11 @@ export class ThreePool extends CurvePool {
   }
 
   handleCoinTransfer(event: any, state: PoolState, log: Log): PoolState {
-    const from = event.args.src;
+    const from = event.args.from;
```
ColonelJ, line 70: I think this is not a correct change, the ABI we are using has `Transfer` event with `src` in it, not `from`.

ColonelJ (approved):

src/dex/curve-v1/pools/3pool.ts
```diff
@@ -118,7 +118,7 @@ export class ThreePool extends CurvePool {
     const rates = this.getRates();
     const i = _.findIndex(
       this.COINS,
-      c => c.toLowerCase() === this.lastTransferedCoin!.toLowerCase(),
+      c => c.toLowerCase() === this.lastTransferredCoin?.toLowerCase(),
```
mwamedacen, line 121: perhaps you can skip findIndex if lastTransferredCoin is undefined, minor though

mwamedacen (approved):

src/dex/curve-v1/pools/3pool.ts
Verisana: Yeah, I agree it is better, but I have already published new version :(

PR #205, opened 2022-10-04
fix: getTopPools curve-v1

Louis-Amas:
Resolves BACK-744.

### DIFF
--- a/src/dex/curve-v1/curve-v1.ts
+++ b/src/dex/curve-v1/curve-v1.ts
@@ -1109,7 +1109,7 @@ export class CurveV1 extends SimpleExchange implements IDex<CurveV1Data> {
             exchange: this.dexKey,
             address: pool.address,
             liquidityUSD: pool.liquidityUSD!,
-            connectorTokens: _.uniqBy(connectorTokens, 'address')
+            connectorTokens: _.uniq(connectorTokens)
               .filter((_token: string) =>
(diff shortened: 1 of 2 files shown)


src/dex/curve-v1/curve-v1.ts
```diff
@@ -1104,10 +1109,11 @@ export class CurveV1 extends SimpleExchange implements IDex<CurveV1Data> {
             exchange: this.dexKey,
             address: pool.address,
             liquidityUSD: pool.liquidityUSD!,
-            connectorTokens: _.uniq(connectorTokens)
-              .filter(
-                (_token: string) =>
-                  _token.toLowerCase() !== _tokenAddress.toLowerCase(),
```
Louis-Amas, line 1110: Logic was inversed here

Louis-Amas:
https://github.com/paraswap/paraswap-pooltracker/commit/2125a68e340e841b74631deb8121039f9544be51#r85841333

erez-paraswap (approved):

src/dex/curve-v1/curve-v1.ts
ColonelJ: It's supposed to be inversed, you are filtering for tokens that DON'T match the input

ColonelJ (changes requested):

PR #207, opened 2022-10-05
Feat/back 743

Louis-Amas:


src/stateful-event-subscriber.ts
```diff
@@ -20,13 +27,85 @@ export abstract class StatefulEventSubscriber<State>
...
+            `${this.parentName}: ${this.name}: found null state in cache generate new one`,
+          );
+          state.state = await this.generateState(blockNumber);
+        }
+        // apply a callback on the state (usefull for initialization for slaves instance)
+        if (cb) {
+          cb(state.state);
+        }
+        this.logger.debug(
+          `${this.parentName}: ${this.name}: found state from cache`,
```
ColonelJ, line 79: Probably shouldn't log this when the state in cache was `null`.

src/stateful-event-subscriber.ts
```diff
@@ -20,13 +27,85 @@ export abstract class StatefulEventSubscriber<State>
...
+        );
+        // set state and the according blockNumber. state.bn can be smaller, greater or equal
+        // to blockNumber
+        this.setState(state.state, state.bn);
```
ColonelJ, line 83: `state.bn` doesn't seem to be necessarily the block number, if the state was generated locally using `blockNumber`.

src/stateful-event-subscriber.ts
```diff
@@ -20,13 +27,85 @@ export abstract class StatefulEventSubscriber<State>
...
+        `${this.parentName}: ${this.name}: cache generating state`,
+      );
+      const state = await this.generateState(blockNumber);
+      this.setState(state, blockNumber);
+    }
+    // always subscribeToLogs
+    this.dexHelper.blockManager.subscribeToLogs(
+      this,
+      this.addressesSubscribed,
+      blockNumber,
```
ColonelJ, line 105: This should always be the block number which was set on the state (could be the one from cache).  Might be simple to just overwrite the `blockNumber` parameter with the value from cache when the cache is used.

ColonelJ (changes requested):
Only reviews on StatefulEventSubscriber.  Need to review again when the bugs are fixed.

src/stateful-event-subscriber.ts
```diff
@@ -134,6 +214,22 @@ export abstract class StatefulEventSubscriber<State>
...
+      this.masterPoolNeeded &&
+      this.state === null &&
+      blockNumberForMissingStateRegen
+    ) {
+      const createNewState = async () => {
+        this.logger.debug(
+          `${this.parentName}: ${this.name}: master generate new state because state is null`,
+        );
+        const state = await this.generateState(blockNumberForMissingStateRegen);
+        this._setState(state, blockNumberForMissingStateRegen);
```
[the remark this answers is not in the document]
Louis-Amas: Not sure why here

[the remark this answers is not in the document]
ColonelJ: It needs to be set in the state history.  In general the result of `generateState` should be used with `setState` not `_setState`.

src/dex/uniswap-v3/uniswap-v3-pool.ts
```diff
@@ -0,0 +1,367 @@
...
+
+    const tickBitmap = {};
+    const ticks = {};
+
+    this._reduceTickBitmap(tickBitmap, _state.tickBitmap);
+    this._reduceTicks(ticks, _state.ticks);
+
+    // Not really a good place to do it, but in order to save RPC requests,
+    // put it here
+    this.poolAddress = _state.pool.toLowerCase();
```
Louis-Amas, line 165: Here

ColonelJ (changes requested):
Few minor things

src/dex/uniswap-v3/uniswap-v3.ts
```diff
@@ -0,0 +1,824 @@
...
+  async getPool(
+    srcAddress: Address,
+    destAddress: Address,
+    fee: bigint,
+    blockNumber: number,
+  ): Promise<UniswapV3EventPool | null> {
+    let pool =
+      this.eventPools[this.getPoolIdentifier(srcAddress, destAddress, fee)];
+    if (pool === undefined) {
+      const [token0, token1] = this._sortTokens(srcAddress, destAddress);
```
Verisana, line 128: I see in types that `pool` may be `UniswapV3EventPool | null`, but it is checked against `undefined`. Don't know if it is intentional, but looks strange

src/dex/uniswap-v3/uniswap-v3.ts
```diff
@@ -0,0 +1,824 @@
...
+        token1,
+        this.logger,
+      );
+
+      try {
+        await pool.initialize(blockNumber, {
+          initCallback: (state: DeepReadonly<PoolState>) => {
+            //really hacky, we need to push poolAddress so that we subscribeToLogs in StatefulEventSubscriber
+            pool!.addressesSubscribed[0] = state.pool;
+            pool!.poolAddress = state.pool;
```
Verisana, line 158: May you please, use type narrowing instead of exclamation mark? I really think that using exclamation marks should be only when it is really needed. Because you are bypassing TS type checking system and make it useless:
```
          if (pool !== null) {
            pool.addressesSubscribed.push(state.pool);
            pool.poolAddress = state.pool;
          } else {
            // Logging for impossible case
          }
```

I saw that you assigned previously new UniswapV3EventPool to `pool` and I don't know too why TS is not picking up automatically that `pool` is not `null` anymore. This change will be helpful if we do some refactor and someone removes accidentally that assignment, then we will catch in logs, that it was broken. But with exclamation mark it will go unnoticed.

src/dex/uniswap-v3/uniswap-v3.ts
Verisana: Ah, remembered it. `undefined` - is for uninitialized pool. In that case, may you please, add comment about that? I think it is not intuitive

src/dex/uniswap-v3/uniswap-v3.ts
```diff
@@ -0,0 +1,824 @@
...
+    ).filter(pool => pool);
+
+    if (pools.length === 0) return [];
+
+    return pools.map(pool =>
+      this.getPoolIdentifier(_srcAddress, _destAddress, pool!.feeCode),
+    );
+  }
+
+  async getStateFromRpc(
```
Verisana, line 236: I think the naming is misleading. It is not state, but `output` or `pricing`

src/dex/uniswap-v3/uniswap-v3.ts
```diff
@@ -0,0 +1,824 @@
...
+          }
+          return acc;
+        },
+        {
+          poolWithState: [] as UniswapV3EventPool[],
+          poolWithoutState: [] as UniswapV3EventPool[],
+        },
+      );
+
+      const rpcResultsPromise = this.getStateFromRpc(
```
Verisana, line 428: If I understood it correctly, we expect that RPC calls should be very low. In that case, we need to add logging, to know how frequently we are using RPC calls.

src/dex/uniswap-v3/uniswap-v3.ts
```diff
@@ -0,0 +1,824 @@
...
+                to.address,
+                pool.feeCodeAsString,
+                _amount.toString(),
+                0, //sqrtPriceLimitX96
+              ]),
+      })),
+    );
+
+    const data = await this.uniswapMulti.methods
+      .multicall(calldata.flat())
```
Verisana, line 290: I don't understand this part. Is it failing for all or can be executed partial? I think it should be executable partially

src/dex/uniswap-v3/uniswap-v3.ts
```diff
@@ -0,0 +1,824 @@
...
+      .call();
+
+    const decode = (j: number): bigint => {
+      if (!data.returnData[j].success) return 0n;
```
Verisana, line 294: Maybe we should log this as well? Because I don't think we expect many calls to fail. But if it is failing, something is wrong.

src/utils.ts
```diff
@@ -182,7 +183,123 @@ export function interpolate(
...
+  (str: string) => BigInt(str),
+);
+
+const casterStringToBigNumber = stringCasterBuilder(
+  PREFIX_BIG_NUMBER,
+  (str: string) => new BigNumber(str),
+);
+
+type TypeSerializer = {
+  checker: (obj: any) => boolean;
```
Verisana, line 252: Maybe something like this is more type safe?
```
  checker: <T>(obj: T) => boolean;
  caster: <IN, OUT>(obj: IN) => OUT;
```

src/utils.ts
```diff
@@ -182,7 +183,123 @@ export function interpolate(
...
+  caster: (obj: any) => any;
+};
+
+export function deepTypecast(obj: any, types: TypeSerializer[]): any {
```
Verisana, line 256: Well, maybe typing so many things will lead to problems and maybe not worth it. Or we can try :)

src/stateful-event-subscriber.ts
```diff
@@ -20,13 +32,112 @@ export abstract class StatefulEventSubscriber<State>
...
+    this.mapKey =
+      `${CACHE_PREFIX}_${this.dexHelper.config.data.network}_${this.parentName}`.toLowerCase();
+    this.cacheName = `${this.mapKey}_${this.name}`.toLowerCase();
+  }
 
   getStateBlockNumber(): Readonly<number> {
     return this.stateBlockNumber;
   }
 
+  //Function which set the initiale state and bounded it to blockNumber
```
Verisana, line 59: `initiale` -> `initial`

src/stateful-event-subscriber.ts
```diff
@@ -20,13 +32,112 @@ export abstract class StatefulEventSubscriber<State>
...
+  //      will also publish a message to cache to tell one master version of dex-lib that this slave
+  //      instance subscribed to a pool from dex this.parentName and name this.name.
+  async initialize(
+    blockNumber: number,
+    options?: InitializeStateOptions<State>,
+  ) {
+    if (options && options.state) {
+      this.setState(options.state, blockNumber);
+    } else {
+      if (this.dexHelper.config.isSlave && this.masterPoolNeeded) {
```
Verisana, line 82: I don't fully understand property `this.masterPoolNeeded`. What is it used for?

src/stateful-event-subscriber.ts
```diff
@@ -134,23 +246,44 @@ export abstract class StatefulEventSubscriber<State>
...
+      blockNumberForMissingStateRegen
+    ) {
+      const createNewState = async () => {
+        this.logger.debug(
+          `${this.parentName}: ${this.name}: master generate new state because state is null`,
+        );
+        const state = await this.generateState(blockNumberForMissingStateRegen);
+        this.setState(state, blockNumberForMissingStateRegen);
+      };
+      createNewState();
```
Verisana, line 263: If we are not awaiting it, I think it worth to catch the error in case of unexpected situation and to not receive unhandled promise rejection

src/dex/uniswap-v3.ts
```diff
@@ -7,6 +7,7 @@ import { SimpleExchange } from './simple-exchange';
 import UniswapV3RouterABI from '../abi/UniswapV3Router.json';
```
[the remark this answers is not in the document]
Louis-Amas: Probably better to keep like that until we work on the correct QuickSwapV3 implem

src/dex/uniswap-v3/scripts/check-pool-bitmap-ranges.ts
```diff
@@ -0,0 +1,238 @@
+// /*
```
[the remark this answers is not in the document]
Louis-Amas: Not compiling because of dex-lib change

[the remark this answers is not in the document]
Louis-Amas: Not worth it to invest time to make it compile properly

src/dex/uniswap-v3/uniswap-v3.ts
```diff
@@ -0,0 +1,824 @@
...
+import { Contract } from 'web3-eth-contract';
+import { AbiItem } from 'web3-utils';
+
+type PoolPairsInfo = {
+  token0: Address;
+  token1: Address;
+  fee: string;
+};
+
+const UNISWAPV3_QUOTE_GASLIMIT = 200_000;
```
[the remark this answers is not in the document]
Louis-Amas: Yes I will keep this one

src/utils.ts
Louis-Amas: We needs to specify the type for all dexes not really worth it

src/stateful-event-subscriber.ts
Louis-Amas: Create a pool in the master version of dex-lib.

src/dex/uniswap-v3/uniswap-v3.ts
Louis-Amas: Ts is not picking up and there is no chance pool is null

src/dex/uniswap-v3.ts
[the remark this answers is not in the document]
Verisana: May you, please, elaborate on reasoning why is it better to keep it here? It is just confusing and raising questions, why we have `UniswapV3` in legacy and new file

src/dex/uniswap-v3/uniswap-v3.ts
Verisana: There is no chance unless someone moves this line for whatever reason somewhere outside:

```
      pool = new UniswapV3EventPool(
        this.dexHelper,
        this.dexKey,
        this.config.stateMulticall,
        this.config.factory,
        fee,
        token0,
        token1,
        this.logger,
      );
```

Verisana (approved):

ColonelJ (approved):

PR #208, opened 2022-10-05
BalancerV1 rewrite with bigint math

ColonelJ:
Rewritten BalancerV1 implementation and tests using the templates.
(Templates themselves were modified to reflect better practices.)
All state now stored using bigint instead of bignumber.js or strings.
New implementation of the Solidity maths, using only bigint.
Pools are now initialized on demand instead of all at startup.
New SOR package version is actually for BalancerV2, so removed it.


mwamedacen (approved):

mwamedacen (approved):

src/dex-helper/dummy-dex-helper.ts
```diff
@@ -100,9 +100,9 @@ class DummyBlockManager implements IBlockManager {
     contractAddress: Address | Address[],
     afterBlockNumber: number,
   ): void {
-    console.log(
-      `Subscribed to logs ${subscriber.name} ${contractAddress} ${afterBlockNumber}`,
-    );
+    // console.log(
+    //   `Subscribed to logs ${subscriber.name} ${contractAddress} ${afterBlockNumber}`,
+    // );
```
Verisana, line 105: These messages were quote useful to understand that subscribe to logs was called in tests

Verisana (approved):

src/dex-helper/dummy-dex-helper.ts
ColonelJ: Was carried over from the previous version that subscribed to all pools at startup (no longer doing that), I have re-enabled this logging now.

PR #214, opened 2022-10-10
Add new ComposableStable pool type.

johngrantuk:
* Adds new ComposableStable pool type
    * ComposableStable has same maths as StablePhantom. The main difference is that ComposableStables have join/exit functions when StablePhantom did not. The difference of note for swaps is ComposableStable must use 'actualSupply' instead of VirtualSupply. 
    * VirtualSupply could be calculated easily whereas actualSupply cannot hence the use of onchain call 'getActualSupply'
* There is a large matic/stMatic pool on [Polygon](https://polygon.balancer.fi/#/pool/0x8159462d255c1d24915cb51ec361f700174cd99400000000000000000000075d) that can be tested against.
* Other aggregators haven't hooked in to this yet.


Verisana:
Thanks @johngrantuk  for preparing this pool :)
Closing this PR in favor of [PR#281](https://github.com/paraswap/paraswap-dex-lib/pull/281) to prepare for merge

PR #220, opened 2022-10-21
Feat/back 760

Louis-Amas:


src/dex/lido.ts
```diff
@@ -1,6 +1,6 @@
 import { Provider } from '@ethersproject/providers';
```
Verisana, line 1: It is not used anywhere

src/dex-helper/irequest-wrapper.ts
```diff
@@ -1,3 +1,45 @@
+export type Method =
```
Verisana, line 1: I think better just to import from `axios`

src/lib/fetcher/fetcher.ts
```diff
@@ -0,0 +1,142 @@
...
+
+export type RequestInfoWithHandler<T> = {
+  info: RequestInfo<T>;
+  handler: (data: T) => void;
+};
+
+export default class Fetcher<T> {
+  private intervalId?: Timeout;
+  private requests: RequestInfoWithHandler<T>[];
+  public lastFetchSucceded: boolean = false;
```
Verisana, line 26: `Succeded` -> `Succeeded`

src/lib/fetcher/fetcher.ts
```diff
@@ -0,0 +1,142 @@
...
+      },
+    );
+    const results = await Promise.all(promises);
+    const failures = results
+      .map((_, i) => i)
+      .filter(i => results[i] instanceof Error);
+
+    failures.forEach(i => {
+      this.logger.warn(
+        `failled polling ${this.requests[i].info.requestOptions.url} ${results[i]}`,
```
Verisana, line 79: `failled` -> `failed`

src/lib/fetcher/fetcher.ts
```diff
@@ -0,0 +1,142 @@
...
+      );
+    });
+
+    if (failures.length === results.length) {
+      this.lastFetchSucceded = false;
```
Verisana, line 84: `Succeded` -> `Succeeded`

src/lib/fetcher/fetcher.ts
```diff
@@ -0,0 +1,142 @@
...
+      this.failedCount += 1;
+    } else {
+      this.lastFetchSucceded = true;
```
Verisana, line 87: `Succeded` -> `Succeeded`

src/lib/fetcher/fetcher.ts
```diff
@@ -0,0 +1,142 @@
...
+    }
+
+    results
+      .map((_, i) => i)
```
Verisana, line 91: Cool thing! I never saw this pattern previously

src/lib/fetcher/fetcher.ts
```diff
@@ -0,0 +1,142 @@
...
+  }
+
+  private getUrls(): string[] {
+    return this.requests.map(el => el.info.requestOptions.url!);
+  }
+
+  startPolling(): void {
+    if (this.intervalId) return;
+
+    this.intervalId = setInterval(() => {
```
Verisana, line 122: Maybe `setInterval` is not best option here, because if you have timeout of 10 sec. and pollling interval of 5 sec. for unresponsive endpoint it will end up making more new tasks than it will be able to resolve, right?

I would stick to recursive `setTimeout`, but need to be carefull to not break the chain of setting new timer.

src/dex-helper/irequest-wrapper.ts
Louis-Amas: Axios is devDependecy

Louis-Amas: If someone want to use dex-lib without axios it should be able to do it I guess

src/lib/fetcher/fetcher.ts
```diff
@@ -0,0 +1,139 @@
...
+          return e as Error;
+        }
+      },
+    );
+    const results = await Promise.all(promises);
+    const failures = results
+      .map((_, i) => i)
+      .filter(i => results[i] instanceof Error);
+
+    failures.forEach(i => {
```
Verisana, line 82: Maybe better to batch this warnings. I am afraid it will just spam our logs

src/lib/fetcher/fetcher.ts
```diff
@@ -0,0 +1,139 @@
...
+    if (failures.length === results.length) {
+      this.lastFetchSucceeded = false;
+      this.failedCount += 1;
+    } else {
+      this.lastFetchSucceeded = true;
+    }
+
+    results
+      .map((_, i) => i)
+      .filter(i => !(results[i] instanceof Error))
```
Verisana, line 97: You don't need to loop twice through all results. But if we expect that `requests.length` is not big, then maybe not worth optimizing

src/lib/fetcher/fetcher.ts
```diff
@@ -0,0 +1,139 @@
...
+  excludedFieldsCaching?: string[];
+};
+
+export type RequestInfoWithHandler<T> = {
+  info: RequestInfo<T>;
+  handler: (data: T) => void;
+};
+
+export default class Fetcher<T> {
+  private intervalId?: Timeout;
```
Verisana, line 24: Name is misleading. It should be changed to `timer` / `timeoutTImer` or whatever

tests/utils-e2e.ts
```diff
@@ -217,6 +223,7 @@ export async function testE2E(
 
   if (paraswap.initializePricing) await paraswap.initializePricing();
 
+  await sleep(5000);
```
Verisana, line 226: Why we sleep here?

tests/utils-e2e.ts
```diff
@@ -19,11 +18,16 @@ import {
...
+  Config,
 } from '../src/types';
 import Erc20ABI from '../src/abi/erc20.json';
 import AugustusABI from '../src/abi/augustus.json';
 import { generateConfig } from '../src/config';
 import { DummyLimitOrderProvider } from '../src/dex-helper';
+import { constructSimpleSDK, SimpleFetchSDK, SimpleSDK } from '@paraswap/sdk';
+import { sleep } from '../tests/utils';
+import axios from 'axios';
+import { SmartToken, StateOverrides } from './smart-tokens';
```
Verisana, line 30: I just wonder, if we really need this `SmartToken` thing. It will require a lot of change. What if we just change `holders` -> to variable names need to override. It should be enough. What do you think, guys? @Louis-Amas @ColonelJ @mwamedacen

tests/constants-e2e.ts
```diff
@@ -1,7 +1,18 @@
...
+  SmartTokenParams,
+  balanceOfFn,
+  allowanceFn,
+  SmartToken,
+} from '../tests/smart-tokens';
+import { Address } from '../src/types';
 import { ETHER_ADDRESS, Network } from '../src/constants';
 
-export const Tokens: { [network: number]: { [symbol: string]: Token } } = {
+export const GENERIC_ADDR1 = '0xbe9317f6711e2da074fe1f168fd9c402bc0a9d1b';
```
Verisana, line 10: Where is this used? I can not find

src/types.ts
```diff
@@ -1,5 +1,5 @@
 import { SwapSide } from './constants';
```
Verisana, line 1: Not used

tests/utils-e2e.ts
Louis-Amas: Btw what you propose is what I started to build, but I found out it's better to create smart token class because it permit to hide also approval and balance for srcToken.

tests/constants-e2e.ts
Louis-Amas: https://github.com/paraswap/paraswap-dex-lib/blob/b26673fbe227a16049d2db3bc14f9a900a155d8c/src/dex/generic-rfq/generic-rfq-e2e.test.ts#L47

tests/utils-e2e.ts
Louis-Amas: We can remove that

src/lib/fetcher/fetcher.ts
Louis-Amas: Yes I know but I think we don't care it's more simple to read

src/dex/generic-rfq/rate-fetcher.ts
```diff
@@ -0,0 +1,387 @@
+import BigNumber from 'bignumber.js';
+import { SwapSide } from 'paraswap-core';
+import { BN_0, BN_1, getBigNumberPow } from '../../bignumber-constants';
+import { IDexHelper } from '../../dex-helper';
+import Fetcher from '../../lib/fetcher/fetcher';
+import { Logger, Address, Token } from '../../types';
+import { OrderInfo } from '../paraswap-limit-orders/types';
+import {
+  BigNumberRate,
+  BigNumberRates,
```
Verisana, line 10: Not used

src/dex/generic-rfq/rate-fetcher.ts
```diff
@@ -0,0 +1,387 @@
...
+  PriceAndAmount,
+  PriceAndAmountBigNumber,
+  Rates,
+  RatesResponse,
+  RFQConfig,
+  RFQFirmRateResponse,
+  RFQPayload,
+} from './types';
+
+export const FIRM_RATE_TIMEOUT_MS = 500;
```
Verisana, line 23: It is quite strict

src/dex/generic-rfq/rate-fetcher.ts
```diff
@@ -0,0 +1,387 @@
...
+        {
+          info: {
+            requestOptions: config.marketConfig.reqParams,
+            caster: this.castMarketResponse.bind(this),
+          },
+          handler: this.handleMarketResponse.bind(this),
+        },
+        config.marketConfig.intervalMs,
+        this.logger,
+      ));
```
Verisana, line 62: Why these expressions in some strange bracekts?

src/dex/generic-rfq/rate-fetcher.ts
```diff
@@ -0,0 +1,387 @@
...
+          `${pair.base.address}_${pair.quote.address}_${SwapSide.BUY}`,
+          this.config.marketConfig.dataTTLS,
+          JSON.stringify(prices.asks),
+        );
+      }
+    }
+  }
+
+  checkHealth(): boolean {
+    return [this.marketFetcher, this.rateFetcher].some(
```
Verisana, line 196: Maybe better `every`?

src/dex/generic-rfq/rate-fetcher.ts
```diff
@@ -0,0 +1,387 @@
...
+      f => f.lastFetchSucceeded,
+    );
+  }
+
+  public async getOrderPrice(
+    srcToken: Token,
+    destToken: Token,
+    side: SwapSide,
+  ): Promise<OrderPriceInfo | null> {
+    // let from = side === SwapSide.SELL ? srcToken : destToken;
```
Verisana, line 206: Commented code

Verisana (commented):
Hopefully it is the last one :)

src/constants.ts
```diff
@@ -1,7 +1,9 @@
-import { Address } from './types';
-export { SwapSide, ContractMethod } from 'paraswap-core';
+export { SwapSide, ContractMethod } from '@paraswap/core';
 
-export const ETHER_ADDRESS = '0xEeeeeEeeeEeEeeEeEeEeeEEEeeeeEeeeeeeeEEeE';
+export const PORT_TEST_SERVER = process.env.TEST_PORT;
```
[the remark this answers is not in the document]
Louis-Amas: I think it can be use for other test later so I prefer to be generic

src/dex/generic-rfq/example-api.test.ts
```diff
@@ -0,0 +1,222 @@
...
+      signature,
+    };
+
+    return res.status(200).json({
+      status: 'accepted',
+      order: orderToPostToApi,
+    });
+  });
+
+  const server = app.listen(parseInt(process.env.TEST_PORT!, 10));
```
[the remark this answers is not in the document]
Louis-Amas: Yes we can but we are on test and the listen will failed

src/dex/generic-rfq/rate-fetcher.ts
Louis-Amas: ?

src/dex/generic-rfq/rate-fetcher.ts
Louis-Amas: Currently it's some on hs

src/dex/generic-rfq/rate-fetcher.ts
Louis-Amas: What do you prefer ?

src/dex/generic-rfq/rate-fetcher.ts
Verisana: But what if one is not functioning? We will not know. Normally both of them should work

src/dex/generic-rfq/rate-fetcher.ts
Verisana: DMed

src/dex/generic-rfq/rate-fetcher.ts
Verisana: Maybe 1 sec. at least? But better to discuss on SU

src/dex/generic-rfq/rate-fetcher.ts
Louis-Amas: I think we should not fail when a third party is failing. Otherwise it will restart api. Btw currently checkHealth is not call anywhere. Because I would like to build on top of initialization service PR

src/dex/generic-rfq/generic-rfq.ts
```diff
@@ -0,0 +1,217 @@
...
+    const unit = unitResults.outputs[0];
+
+    const outputs = this.calcOutsFromAmounts(
+      _amountsInBN,
+      getBigNumberPow(outDecimals),
+      orderPricesInfo.rates,
+    );
+    return [
+      {
+        gasCost: Number(BigInt(outputs.ordersCount) * ONE_ORDER_GASCOST),
```
ColonelJ, line 173: There seems to be some confusion here, the order book is only used to define the price curve, there is no gas cost associated with how many entries we went through.  The firm rate is always returned as one order from the MM.

ColonelJ (changes requested):
Generic RFQ using AugustusRFQ should not use `txOrigin`, we don't care about that, should instead use `userAddress`.  I don't think we're passing it through transaction preprocessing at the moment.

PR #223, opened 2022-11-01
integrate babydogeswap

itdev-1210:
Website: https://babydogeswap.com/
Docs: https://baby-doge-coin.gitbook.io/babydoge-docs/


itdev-1210:
Hi, @ColonelJ 
Can you please review our PR?

This is for BabyDogeSwap.

ColonelJ:
Hi, this looks like a fork of UniswapV2, instead of copying the code you can simply add your DEX to the existing config.ts of uniswap-v2, you can see we already have a lot of forks listed.  If there's anything slightly different/special in your smart contract implementation which isn't already covered by the config (e.g. individual pool fees), you could instead extend the UniswapV2 class into another file inside uniswap-v2 with the config inside (there are some examples of that too).  Obviously, if you can add your DEX with only config that is far preferable for us.  Thanks!

itdev-1210:
Hi, Thank you for the feedback.
We are forking Uniswap-V2.
so, just only add config like pancakeswap?

itdev-1210:
Remake.

PR #225, opened 2022-11-07
Feature/swaap v1

Louis-Amas:


src/dex/swaap-v1/swaap-v1.ts
```diff
@@ -0,0 +1,621 @@
...
+    );
+
+    if (srcTokenAddress === destTokenAddress) return null;
+    const currentTimestamp =
+      _currentTimeStamp || BigInt(Math.floor(Date.now() / 1000)); // the on-chain pricing logic is not too much sensitive to that value
+    return await Promise.all(
+      allowedPools.map(async poolAddress => {
+        let state = this.eventPools[poolAddress].getState(blockNumber);
+        if (!state) {
+          state = await this.eventPools[poolAddress].generateState(blockNumber);
```
ColonelJ, line 582: Should call `setState` with the generated state so that it can possibly be used again.

src/dex/swaap-v1/swaap-v1.ts
```diff
@@ -0,0 +1,621 @@
...
+      this.dexHelper,
+    );
+    const state = await pool.generateState(blockNumber);
+    pool.setState(state, blockNumber);
+
+    this.dexHelper.blockManager.subscribeToLogs(
+      pool,
+      pool.addressSubscribers,
+      blockNumber,
+    );
```
ColonelJ, line 119: Can migrate generate/set state and subscribe to logs to using new `initialize` function instead.  It looks like `addressSubscribers` needs renamed everywhere for that to work.

ColonelJ (changes requested):

ColonelJ (approved):

Verisana (approved):

PR #237, opened 2022-11-17
feat: curve-v1: disable fee on transfer for stETH

Louis-Amas:
Resolves BACK-777.

### DIFF
--- a/src/dex/curve-v1/curve-v1.ts
+++ b/src/dex/curve-v1/curve-v1.ts
@@ -54,6 +54,7 @@ import {
   getBigIntPow,
   getDexKeysWithNetwork,
   interpolate,
+  isDestTokenTransferFeeToBeExchanged,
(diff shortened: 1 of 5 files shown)


Verisana (approved):

src/dex/curve-v1/curve-v1.ts
```diff
@@ -294,8 +306,8 @@ export class CurveV1 extends SimpleExchange implements IDex<CurveV1Data> {
   }
 
   getPoolConfigs(from: Token, to: Token) {
-    const fromAddress = from.address.toLowerCase();
-    const toAddress = to.address.toLowerCase();
+    const fromAddress = from.address;
+    const toAddress = to.address;
```
ColonelJ, line 310: This change looks wrong...

ColonelJ (changes requested):

PR #245, opened 2022-11-29
Feat/back 786

Louis-Amas:


src/lib/tokens/balancer-fetcher.ts
```diff
@@ -247,6 +246,7 @@ export const decodeBalanceAndAllowanceMultiResult = (
 export const getBalances = async (
   multiv2: MultiWrapper,
   reqs: BalanceRequest[],
+  blockNumber: number | string = 'latest',
```
Verisana, line 249: I think this is type safer:
`  blockNumber: number | 'latest' = 'latest',`

src/lib/generics-events-subscribers/erc20-event-subscriber-factory.ts
```diff
@@ -0,0 +1,18 @@
+import { IDexHelper } from '../../dex-helper';
+import { Address } from '../../types';
+import { ERC20EventSubscriber } from './erc20-event-subscriber';
+
+const subscriberMap: Record<Address, ERC20EventSubscriber> = {};
```
Verisana, line 5: I don't understand why wee need this mapping? Just to keep instance running after it was returned?

src/lib/generics-events-subscribers/erc20-event-subscriber.ts
```diff
@@ -0,0 +1,202 @@
...
+import { ERC20Event, ERC20StateMap, WrappedEvent } from './types';
+
+export const handleTransferEvent = (
+  event: LogDescription,
+  state: ERC20StateMap,
+) => {
+  const erc20Transfer = decodeERC20Transfer(event);
+
+  if (erc20Transfer.from in state) {
+    state = _.cloneDeep(state);
```
Verisana, line 28: It is not allowed to directly modify state without making cloneDeep?
I remember we were doing that, but maybe it is changed since last time I saw it

Verisana (approved):

src/lib/tokens/balancer-fetcher.ts
Louis-Amas: True

src/lib/generics-events-subscribers/erc20-event-subscriber-factory.ts
Louis-Amas: If some others dexes want to the same erc20 tracking

src/lib/generics-events-subscribers/erc20-event-subscriber.ts
Louis-Amas: No it's not it will corrupt in case of reorg

PR #255, opened 2022-12-08
feat: enable token tracking only on master dex-lib instance

Louis-Amas:
Resolves BACK-824.

### DIFF
--- a/src/dex/uniswap-v3/uniswap-v3-pool.ts
+++ b/src/dex/uniswap-v3/uniswap-v3-pool.ts
@@ -124,7 +124,7 @@ export class UniswapV3EventPool extends StatefulEventSubscriber<PoolState> {
     // only if the super call succeed
 
     const initPromises = [];
-    if (!this.token0sub.isInitialized) {
+    if (!this.token0sub.isInitialized && !this.dexHelper.config.isSlave) {
       initPromises.push(
         this.token0sub.initialize(blockNumber, {
           state: {},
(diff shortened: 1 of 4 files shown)


src/dex/uniswap-v3/uniswap-v3-pool.ts
```diff
@@ -124,15 +124,15 @@ export class UniswapV3EventPool extends StatefulEventSubscriber<PoolState> {
     // only if the super call succeed
 
     const initPromises = [];
-    if (!this.token0sub.isInitialized) {
+    if (!this.token0sub.isInitialized && !this.dexHelper.config.isSlave) {
```
Verisana, line 127: I don't understand this part fully. Shouldn't we don't create `token0sub` and `token1sub` instances if it is slave version? It will just waste resources processing logs? Or if not subscribed, it doesn't do any work?

But later this thing will be executed:

```
    await Promise.all([
      this.token0sub.subscribeToWalletBalanceChange(
        this.poolAddress,
        blockNumber,
      ),
      this.token1sub.subscribeToWalletBalanceChange(
        this.poolAddress,
        blockNumber,
      ),
    ]);
```
Maybe exit earlier if it is not master version?

Verisana (commented):
I am not sure I found where you use that cached values as slave?

src/dex/uniswap-v3/uniswap-v3-pool.ts
Louis-Amas: We want to access the getBalance funciton that check redis

Verisana (approved):

Verisana (approved):

mwamedacen (approved):

PR #281, opened 2023-01-13
BACK-872: Balancer-V2 Generic Multihop with ComposableStable pools

mwamedacen:


Velenir (approved):
Monumental work :clap:

alexshchur (approved):

PR #282, opened 2023-01-16
fix/univ3-inefficiencies

mwamedacen:
resolves BACK-886

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -341,4 +341,5 @@ export const isDestTokenTransferFeeToBeExchanged = (
   return !!(transferFees.destFee || transferFees.destDexFee);
 };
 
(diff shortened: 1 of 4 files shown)


Verisana (approved):

src/utils.ts
```diff
@@ -340,3 +340,5 @@ export const isDestTokenTransferFeeToBeExchanged = (
 ) => {
   return !!(transferFees.destFee || transferFees.destDexFee);
 };
+
+export const isTruthy = <T>(x: T | undefined | null): x is T => !!x;
```
Velenir, line 344: You need all the falsy values here, including `false | 0 | ""`

mwamedacen: good catch  🫡

mounibec (approved):

Verisana (approved):

PR #284, opened 2023-01-17
Adding Hasflow DEX to Paraswap

mib-hashflow:
This PR adds a Paraswap integration for Hashflow using the `@hashflow/taker-js` library.

I've filled out both the E2E and integration tests. The simple swap E2E tests pass, but multiSwap fails since we don't have Adapters yet.

The only remaining question will be: _"how should we pass the Hashflow authentication key securely?"_

### DIFF
--- a/package.json
+++ b/package.json
@@ -50,6 +50,8 @@
   },
   "dependencies": {
     "@0x/utils": "^4.5.2",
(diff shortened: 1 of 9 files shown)


mib-hashflow:
Merged by Paraswap team

PR #294, opened 2023-02-01
BACK-905: Fix BalancerV2 ComposableStable issue

Verisana:


src/dex/balancer-v2/balancer-v2-integration.test.ts
```diff
@@ -397,3 +358,82 @@ describe('BalancerV2', function () {
...
+
+    const poolPrices = await beetsFi.getPricesVolume(
+      srcToken,
+      destToken,
+      amounts,
+      SwapSide.SELL,
+      blockNumber,
+      pools,
+    );
+    console.log('Pool Prices: ', poolPrices);
```
aburkut-ledger, line 432: @Verisana Are you sure we need this `console.log` here?

paraschiv-razvan-toptal (approved):

skypper (approved):

skypper (approved):

src/dex/balancer-v2/balancer-v2-integration.test.ts
Verisana: Not really. Personally, I don't see much usefulness in many of this `console.logs()` because I more tend to debug inside code rather than printing values, but for some of our collegues, it was useful, so we kept. We can remove, if we see that we are not using that. But I prefer it to do as separate task of improving our tests

PR #302, opened 2023-02-10
BACK-901: Add AaveV3 on Mainnet in configs

skypper:


src/dex/aave-v3/aave-v3.ts
```diff
@@ -17,12 +17,16 @@ import { IDexHelper } from '../../dex-helper/idex-helper';
...
+import { getATokenIfAaveV3Pair, setTokensOnNetwork } from './tokens';
 
 import WETH_GATEWAY_ABI from '../../abi/aave-v3-weth-gateway.json';
 import POOL_ABI from '../../abi/AaveV3_lending_pool.json';
+import { fetchTokenList } from './utils';
 
 const REF_CODE = 1;
+export const TOKEN_LIST_CACHE_KEY = 'token-list';
+const TOKEN_LIST_TTL_SECONDS = 86400; // 1 day
+const TOKEN_LIST_LOCAL_TTL_SECONDS = 10800; // 3 hours
```
Verisana, line 29: I think if you put it like: `24 * 60 * 60` and `3 * 60 * 60`, it is easier to validate

src/dex/aave-v3/utils.ts
```diff
@@ -0,0 +1,156 @@
...
+      'address variableDebtTokenAddress',
+      'address interestRateStrategyAddress',
+      'uint128 accruedToTreasury',
+      'uint128 unbacked',
+      'uint128 isolationModeTotalDebt',
+    ],
+    toDecode,
+  );
+
+  return decoded.aTokenAddress;
```
Verisana, line 47: I think better to return already `toLowercase()` from here

src/dex/aave-v3/utils.ts
```diff
@@ -0,0 +1,156 @@
...
+
+  const results = await multiWrapper.tryAggregate<string | number>(
+    false,
+    calls,
+    blockNumber,
+  );
+
+  let tokenList = reservesList.map((tokenAddress: string, i: number) => ({
+    address: tokenAddress as string,
+    aAddress: results[i].returnData as string,
```
Verisana, line 91: What concerns me, is in case of error you return 'error' as string and pass it here. Maybe better to throw in case of error? line 23

src/dex/aave-v3/utils.ts
```diff
@@ -0,0 +1,156 @@
...
+  }
+  for (const proxyAddress of reservesList) {
+    calls.push({
+      target: proxyAddress,
+      callData: erc20Interface.encodeFunctionData('decimals'),
+      decodeFunction: uint8ToNumber,
+    });
+  }
+
+  const results = await multiWrapper.tryAggregate<string | number>(
```
Verisana, line 83: If we don't expect any call to fail, better to use `aggregate`. It is more efficient and will fail explicitly. But only if all calls must pass

src/dex/aave-v3/utils.ts
```diff
@@ -0,0 +1,156 @@
...
+  let calls: MultiCallParams<any>[] = [];
+
+  for (const proxyAddress of aTokens) {
+    calls.push({
+      target: proxyAddress,
+      callData: erc20Interface.encodeFunctionData('symbol', []),
+      decodeFunction: stringDecode,
+    });
+  }
+  const results = await multiWrapper.tryAggregate<string>(
```
Verisana, line 113: If we don't expect any call to fail, better to use `aggregate`. It is more efficient and will fail explicitly. But only if all calls must pass

src/dex/aave-v3/aave-v3.ts
skypper: Done.

src/dex/aave-v3/utils.ts
skypper: Done.

src/dex/aave-v3/utils.ts
skypper: Done.

src/dex/aave-v3/utils.ts
skypper: There is no reason any call should fail actually.

Done.

src/dex/aave-v3/utils.ts
skypper: There is no reason any call should fail actually.

Done.

Verisana (approved):

aburkut (approved):

src/dex/aave-v3/aave-v3-e2e.test.ts
```diff
@@ -1,23 +1,59 @@
...
+import POOL_ABI from '../../abi/AaveV3_lending_pool.json';
+import ERC20ABI from '../../abi/erc20.json';
+import { Config } from './config';
 
 jest.setTimeout(1000 * 60 * 3);
 
 describe('AaveV3 E2E', () => {
   const dexKey = 'AaveV3';
 
+  beforeAll(async () => {
```
mwamedacen, line 22: Isn't this redundant with the [initializePricing](https://github.com/paraswap/paraswap-dex-lib/blob/master/tests/utils-e2e.ts#L229) call in testE2E ?

src/dex/aave-v3/aave-v3-e2e.test.ts
```diff
@@ -311,27 +314,20 @@ describe('AaveV3 E2E', () => {
         SwapSide.SELL,
         [
           ContractMethod.simpleSwap,
-          ContractMethod.multiSwap,
-          ContractMethod.megaSwap,
+          // ContractMethod.multiSwap,
```
mwamedacen, line 317: please uncomment

src/dex/aave-v3/utils.ts
```diff
@@ -0,0 +1,153 @@
...
+
+  if (!isSuccess || toDecode === '0x') {
+    throw new Error(
+      'Could not extract value from struct DataTypes.ReserveData.',
+    );
+  }
+
+  const decoded = defaultAbiCoder.decode(
+    [
+      'tuple(uint256) configuration',
```
mwamedacen, line 30: better to use common [ABI file](https://github.com/paraswap/paraswap-dex-lib/blob/478d6b231bc55e569adea4cc7baf7a50d8cc305b/src/abi/AaveV3_lending_pool.json#L994) and decode via [Interface](https://docs.ethers.org/v5/api/utils/abi/interface/)

mwamedacen (changes requested):

src/dex/aave-v3/utils.ts
Verisana: I am not sure it is possible to pass custom `Interface` to decoder function. Or it would be something predefined in Utils with closure and will be looking ugly.
In other places we always use this approach of decoding with custom types :)

src/dex/aave-v3/aave-v3-e2e.test.ts
skypper: Due to the way the tests are written this data is required to call getTokenFromASymbol(network, **pair.aTokenSymbol**) in order to pass parameters to testE2E(). If I removed these calls I would have to add hardcoded data in the tests.

The token list is cached, so it shouldn't affect test's performance.

src/dex/aave-v3/aave-v3-e2e.test.ts
skypper: Done.

mwamedacen (approved):

PR #349, opened 2023-03-26
more friendly deadline logic

mwamedacen:


src/transaction-builder.ts
```diff
@@ -43,6 +47,17 @@ export class TransactionBuilder {
     beneficiary?: Address;
     onlyParams?: boolean;
   }) {
+    if (deadline) {
+      const globalDeadline = +deadline;
+      if (!isNaN(globalDeadline)) {
```
Verisana, line 52: What if `globalDeadline` isNan? I think in that case it will continue execution, which may be not right

Verisana (commented):
One small thing

src/transaction-builder.ts
mwamedacen: I agree better fire an error

Verisana:
Merging in this [PR#365](https://github.com/paraswap/paraswap-dex-lib/pull/365)

PR #380, opened 2023-05-09
Maverick-v1 update subgraph URL

gte620v:


Verisana:
Already merged as discussed in Telegram

PR #381, opened 2023-05-10
Back 1073 Implement TraderJoeV2.1 support

paraschiv-razvan-toptal:


skypper (approved):

aburkut (approved):

src/dex/trader-joe-v2.1.ts
```diff
@@ -0,0 +1,155 @@
...
+  }
+
+  getSimpleParam(
+    srcToken: Address,
+    destToken: Address,
+    srcAmount: NumberAsString,
+    destAmount: NumberAsString,
+    data: TraderJoeV2Data,
+    side: SwapSide,
+  ): AsyncOrSync<SimpleExchangeParam> {
```
aburkut, line 118: @paraschiv-razvan-toptal Why do we need AsyncOrSync here?

Verisana: I think simpleParams should be sync. All async operations should happen in preprocessTransaction if needed

PR #421, opened 2023-06-20
BACK-1153: add slippage tolerance for SwaapV2 and Hashflow when restricting

aburkut:


src/dex/hashflow/hashflow.ts
```diff
@@ -657,30 +662,60 @@ export class Hashflow extends SimpleExchange implements IDex<HashflowData> {
...
+      ) {
+        isTooStrictSlippage = true;
+      }
+
+      if (isFailOnSlippage && isTooStrictSlippage) {
+        this.logger.warn(
+          `${this.dexKey}-${this.network}: Market Maker ${mm} failed to build transaction on side ${side} with too strict slippage. Skipping restriction`,
+        );
+      } else if (isFailOnSlippage && !isTooStrictSlippage) {
+        throw new SlippageCheckError(slippageErrorMessage);
```
Verisana, line 716: I am not sure this is correct. If you fail on slippage, you should throw, but restrict only if slippage is not too much. But this code won't throw if slippage is too strict

Verisana (changes requested):

src/dex/hashflow/hashflow.ts
aburkut: @Verisana updated

Verisana (approved):

paraschiv-razvan-toptal (approved):

PR #424, opened 2023-06-22
Fix/fix pos when no partner fee

mwamedacen:
resolves BACK-1158


src/router/buy.ts
```diff
@@ -61,6 +61,13 @@ export class Buy extends PayloadEncoder implements IRouter<BuyParam> {
       minMaxAmount,
       priceRoute.srcAmount,
     );
+
```
paraschiv-razvan-toptal, line 65: Maybe this code sequence may be extracted in a separate method

Verisana (approved):

PR #432, opened 2023-07-07
Fix/camelot multipath

mwamedacen:
resolves BACK-1181

### DIFF
--- a/src/dex/camelot/camelot.ts
+++ b/src/dex/camelot/camelot.ts
@@ -1,6 +1,5 @@
 import {
   DEST_TOKEN_PARASWAP_TRANSFERS,
-  ETHER_ADDRESS,
   Network,
   NULL_ADDRESS,
   SRC_TOKEN_PARASWAP_TRANSFERS,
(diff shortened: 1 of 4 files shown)


Verisana (approved):

aburkut (approved):

src/dex/camelot/camelot.ts
```diff
@@ -729,21 +717,22 @@ export class Camelot
...
     const payload = this.abiCoder.encodeParameter(
       {
         ParentStruct: {
           weth: 'address',
           pools: 'uint256[]',
+          isFeeTokenInRoute: 'bool',
         },
       },
-      { pools, weth },
+      { pools, weth, isFeeTokenInRoute: data.isFeeTokenInRoute },
```
mwamedacen, line 735: because this was missing, decoding was completely erroneous

src/dex/camelot/config.ts
```diff
@@ -19,7 +19,6 @@ export const CamelotConfig: DexConfigMap<DexParams> = {
 
 export const Adapters: Record<number, AdapterMappings> = {
   [Network.ARBITRUM]: {
-    [SwapSide.SELL]: [{ name: 'ArbitrumAdapter02', index: 1 }],
-    [SwapSide.BUY]: [{ name: 'ArbitrumBuyAdapter', index: 1 }],
+    [SwapSide.SELL]: [{ name: 'ArbitrumAdapter02', index: 5 }],
```
mwamedacen, line 22: We were using the same Adapter as for Solidly but as it calls `pool.stable()`  which fails because the function name in Camelot pair is` stableSwap()`

mwamedacen: new adapter: https://arbiscan.io/address/0xD1F70c98a78d48A93F0B4dDa49057469dc5aC126#code

src/dex/camelot/camelot.ts
mwamedacen: as the isFeeTokenInRoute bit was confused with the next bit which lead to take a different contract path

PR #433, opened 2023-07-07
BACK-1182: Persist UniswapV2 & Solidly pairs in distant and local cache

aburkut:


paraschiv-razvan (approved):

src/dex/uniswap-v2/uniswap-v2.ts
```diff
@@ -271,13 +274,19 @@ export class UniswapV2
       callDecoder,
       this.decoderIface,
     );
-    pair.pool.addressesSubscribed.push(pair.exchange!);
+    this.pairToPoolMap[this.getKeyForPair(token0, token1)].addressesSubscribed.push(pair.exchange!);
```
skypper, line 277: Why not save `this.getKeyForPair(token0, token1)` to a variable?

aburkut: @skypper updated

src/dex/uniswap-v2/uniswap-v2.ts
```diff
@@ -340,7 +366,15 @@ export class UniswapV2
     } else {
       pair = { token0, token1, exchange };
     }
-    this.pairs[key] = pair;
+
+    this.dexHelper.cache.setexAndCacheLocally(
```
skypper, line 370: I think it should not cache if `exchange` is NULL_ADDRESS because we already know token0 and token1 and it just consumed cache memory.

skypper: Same for `findSolidlyPair`.

skypper: Cache only the `exchange` or its absence as discussed.

aburkut: @skypper updated

skypper (approved):

mwamedacen (approved):

PR #437, opened 2023-07-12
Fix bids to asks

PawelK-DevBrother:
For buy, ask price should be used


Verisana (approved):

skypper:
There are cases when for. buy if there are no ask prices it uses bids as of here:
https://github.com/paraswap/paraswap-dex-lib/blob/master/src/dex/generic-rfq/rate-fetcher.ts#L340

mwamedacen (approved):

skypper (approved):

PR #449, opened 2023-07-21
implement fallback to multi step state generation + lazy state polling

mwamedacen:
resolves BACK-1202

### DIFF
--- a/src/dex/algebra/algebra-pool.ts
+++ b/src/dex/algebra/algebra-pool.ts
@@ -174,7 +174,7 @@ export class AlgebraEventPool extends StatefulEventSubscriber<PoolState> {
     return TICK_BITMAP_TO_USE + TICK_BITMAP_BUFFER;
   }
 
-  private async _fetchStateSingleStep(
+  private async _fetchPoolStateSingleStep(
     blockNumber: number,
   ): Promise<[bigint, bigint, DecodedStateMultiCallResultWithRelativeBitmaps]> {
     const callData: MultiCallParams<
@@ -231,7 +231,7 @@ export class AlgebraEventPool extends StatefulEventSubscriber<PoolState> {
     return [balance0, balance1, _state];
   }
 
-  private async _getStateRequestCallDataMultiStep(
+  private async _fetchPoolStateMultiStep(
     blockNumber: number,
   ): Promise<[bigint, bigint, DecodedStateMultiCallResultWithRelativeBitmaps]> {
     const balancesAndGlobalStateCalldata: MultiCallParams<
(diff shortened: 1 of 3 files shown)


src/dex/algebra/algebra-pool.ts
```diff
@@ -227,6 +228,123 @@ export class AlgebraEventPool extends StatefulEventSubscriber<PoolState> {
...
+    const _state: DecodedStateMultiCallResultWithRelativeBitmaps = {
+      ..._stateWithoutTicksAndTickBitmap,
+      tickBitmap: ticksAndBitMaps.flatMap(v => v[0]),
+      ticks: ticksAndBitMaps.flatMap(v => v[1]),
+    };
+
+    return [balance0, balance1, _state];
+  }
+
+  async _fetchInitStateMultiStrategies(
```
skypper, line 329: If multicall is consistently failing for complex eth_call RPCs on zkEVM then shouldn't it always pick the multi step strategy for zkEVM?

src/dex/algebra/algebra-pool.ts
```diff
@@ -227,6 +228,123 @@ export class AlgebraEventPool extends StatefulEventSubscriber<PoolState> {
...
+
+      if (this.dexHelper.config.data.network != Network.ZKEVM) throw e;
+
+      return this._fetchPoolStateMultiStep(blockNumber);
+    }
+  }
+
+  async generateState(blockNumber: number): Promise<Readonly<PoolState>> {
+    const [balance0, balance1, _state] =
+      await this._fetchInitStateMultiStrategies(blockNumber);
```
Verisana, line 346: As far as I understand, this function might fail, if we are on zkEVM and multi step multicall fails for some reason. In that case, I don't see any error handling for generate state. Maybe better to consider try catch for that situation [here](https://github.com/paraswap/paraswap-dex-lib/blob/bcac90b3f3c3817c8461eeb7573284f024ff246c/src/dex/algebra/algebra-pool.ts#L114) or [here](https://github.com/paraswap/paraswap-dex-lib/blob/bcac90b3f3c3817c8461eeb7573284f024ff246c/src/dex/algebra/algebra-pool.ts#L340) or [here](https://github.com/paraswap/paraswap-dex-lib/blob/bcac90b3f3c3817c8461eeb7573284f024ff246c/src/dex/algebra/algebra-pool.ts#L303) ?

mwamedacen: This case the status quo basically. When generateState fails, it's caught and state stays at init value (null)

Verisana: Ah, thanks. Found it. Here it is caught:
https://github.com/paraswap/paraswap-api/blob/37f6c026612fc36f834554a22ef363e8ae1c99f1/src/lib/block-manager.ts#L392

Verisana (approved):

skypper (approved):

src/dex/algebra/algebra.ts
```diff
@@ -419,29 +436,43 @@ export class Algebra extends SimpleExchange implements IDex<AlgebraData> {
...
+      let state = pool.getState(blockNumber);
 
       if (state === null) {
         if (this.network === Network.ZKEVM) {
-          if (!pool.initFailed) {
+          if (pool.initFailed) return null;
+
+          if (
+            blockNumber - pool.getStateBlockNumber() <=
+            MAX_STALE_STALE_BLOCK_AGE[this.network]
```
mwamedacen, line 447: Typo ^^’

paraschiv-razvan (approved):

PR #456, opened 2023-08-03
Back 1219/algebra v19 camelotv3 eventbased

mwamedacen:
Bunch of code but the actual diff between algebra v1.1 and v1.9 is that 
- tickSpacing is now configurable
- different fees are applied depending on side


aburkut (approved):

paraschiv-razvan (approved):

src/dex/algebra/algebra-e2e.test.ts
```diff
@@ -181,4 +181,28 @@ describe('Algebra', () => {
       );
     });
   });
+
+  describe('CamelotV3', () => {
+    const dexKey = 'ZyberSwapV3';
```
Verisana, line 186: Why 'ZyberSwapV3'? I can not find USDC/DAI for CamelotV3

mwamedacen: obviously wrong , solved

PR #366, opened 2023-04-12
Add Morphex (GMX fork on Fantom)

pbnather:
Hi, this is initial PR for Morphex (GMX fork on Fantom). 

Morphex inherits from GMX, and all tests are based on GMX ones. 

Integration tests and events tests work properly, however in E2E tests only SimpleSwap works. 
This is because there is no proper adapter on Fantom network for Morphex.

Not sure how to go forward without the adapter, so I reached out in Discord, and created this PR.

Best regards, 
pbnather

PR #473, opened 2023-08-21
Add Morphex (GMX fork)

pbnather:
This PR overwrites changes in https://github.com/paraswap/paraswap-dex-lib/pull/366

After Multichain hack on Fantom, contracts were changed to use layer zero and axelar tokens.

Simple tests work, but Morphex still needs an Adapter written by Paraswap team. 

EDIT: Added BSC config and tests too.


pbnather:
Added BSC config and tests too.

PR #505, opened 2023-09-18
feat: integrate Reservoir DEX

xenide:
## Overview

Reservoir is an AMM with 2 curve types: 
1. Constant Product (x * y = k) 
2. Stable (like curve v2's)

## Documentation 

- [Docs](https://docs.reservoir.fi) 

## Contract addresses 

- AVAX Mainnet [contract addresses](https://docs.reservoir.fi/technical/contract-addresses/avalanche)

## Outstanding issues 

- some events test failing due to engine not providing emitted events to `processLog`
  - in my investigation, I noticed is that it only the latest event emitted by our target (multiple copies of the only one) is being forwarded to processLog by `processBlockLogs`. The problem seems to be somewhere up the call chain from `processBlockLogs`
- e2e tests failing due to no adapters in place for our DEX at the moment


xenide:
can I trouble @aburkut to review this?

PR #515, opened 2023-09-27
BACK-1307: integrate Dexalot on Avalanche chain

aburkut:
Original PR - https://github.com/paraswap/paraswap-dex-lib/pull/485


JackieJoo (approved):

src/dex/dexalot/dexalot.ts
```diff
@@ -0,0 +1,935 @@
...
+      }
+    }
+    return null;
+  }
+
+  getIdentifier(baseSymbol: string, quoteSymbol: string) {
+    // return `${this.dexKey}_${baseSymbol}_${quoteSymbol}`.toLowerCase();
+    // To disable Dexalot in multi/mega routes
+    return this.dexKey;
+  }
```
Verisana, line 203: I am not sure this solution works fully. Yeah, it works in default case, but if `includeContractMethods=megaSwap` is provided, I still see Dexalot

src/dex/dexalot/dexalot.ts
```diff
@@ -0,0 +1,935 @@
...
+      const normalizedDestToken = this.normalizeToken(destToken);
+
+      this.tokensMap = (await this.getCachedTokens()) || {};
+      if (normalizedSrcToken.address === normalizedDestToken.address) {
+        return null;
+      }
+
+      if (limitPools && limitPools.length !== 1) {
+        return null;
+      }
```
Verisana, line 382: I am not sure this the correct way of handling limitPools. We need to check the content, not only number of elements.

src/dex/dexalot/dexalot.ts
```diff
@@ -0,0 +1,935 @@
...
+      this.logger.warn(
+        `${this.dexKey}-${this.network}: blacklisted TX Origin address '${options.txOrigin}' trying to build a transaction. Bailing...`,
+      );
+      throw new Error(
+        `${this.dexKey}-${
+          this.network
+        }: user=${options.txOrigin.toLowerCase()} is blacklisted`,
+      );
+    }
+    if (await this.isRateLimited(options.txOrigin)) {
```
Verisana, line 480: If we have notion of rate limiting on Dexalot, we might want to adjust API to act accordingly and hide Dexalot for user, if he is rate limited. Otherwise, I am afraid we might have quite a lot failures.

src/dex/dexalot/dexalot.ts
```diff
@@ -0,0 +1,935 @@
...
+    );
+  }
+
+  async getTopPoolsForToken(
+    tokenAddress: Address,
+    limit: number,
+  ): Promise<PoolLiquidity[]> {
+    if (await this.isRestricted()) {
+      return [];
+    }
```
Verisana, line 867: I don't think we should skip here if it is restricted. Even if Dexalot is restricted we should still serve good connectors

aburkut: @Verisana updated

src/dex/dexalot/dexalot.ts
aburkut: @Verisana updated

src/dex/dexalot/dexalot.ts
aburkut: not actual anymore

src/dex/dexalot/dexalot.ts
```diff
@@ -0,0 +1,998 @@
...
+    amounts: bigint[],
+    side: SwapSide,
+    blockNumber: number,
+    limitPools?: string[],
+    transferFees?: TransferFeeParams,
+  ): Promise<null | ExchangePrices<DexalotData>> {
+    try {
+      if (await this.isRestricted()) {
+        return null;
+      }
```
Verisana, line 423: I am not sure this logic should be on dex level. We check it somewhere outside, if I am not mistaken:
https://github.com/paraswap/paraswap-api/blob/75554fd9d3eca5af1d4922998ad79504d6bda031/src/lib/hopper/chain-manager.ts#L973-L978

The same, btw, with SwaapV2. Maybe good to keep before RFQ refactor task

JackieJoo (approved):

Verisana (approved):

PR #563, opened 2023-11-06
BACK-1368: non-existent pool caching bug

JackieJoo:


aburkut (approved):

src/dex/algebra/algebra.ts
```diff
@@ -165,7 +165,7 @@ export class Algebra extends SimpleExchange implements IDex<AlgebraData> {
     token0,
     token1,
   }) => {
-    const logPrefix = '[Algebra.onPoolCreatedDeleteFromNonExistingSet]';
+    const logPrefix = `[${this.dexKey}.onPoolCreatedDeleteFromNonExistingSet]`;
```
mwamedacen, line 168: `dexKey` is redundant, all logs are already namespaced with `dexKey` and `chainId`

JackieJoo: removed

src/dex/algebra/algebra.ts
```diff
@@ -176,13 +176,19 @@ export class Algebra extends SimpleExchange implements IDex<AlgebraData> {
...
       );
       // delete pool record from set
-      await this.dexHelper.cache.zrem(this.notExistingPoolSetKey, [poolKey]);
+      const result = await this.dexHelper.cache.zrem(
+        this.notExistingPoolSetKey,
+        [poolKey],
+      );
+      this.logger.info(
+        `${logPrefix} delete pool from not existing set=${this.notExistingPoolSetKey}; key=${poolKey}; result: ${result}`,
+      );
```
KanievskyiDanylo, line 188: Does it make sense to `log` it, if we will log it anyway in case of error?

JackieJoo: Here it's mostly to log the `result`, if it's > 0, means we actually deleted smth, if it's 0, means there wasn't any pool pair by that key

Error is thrown when we try to run zrem against non-sets

KanievskyiDanylo:
I think, it makes sense to add extra delete operation [here](https://github.com/paraswap/paraswap-dex-lib/blob/feat/BACK-1368/src/dex/algebra/algebra.ts#L321):
```ts
    if (pool !== null) {
     this.dexHelper.cache.zrem(
        this.notExistingPoolSetKey,
        [poolKey],
      );
      const allEventPools = Object.values(this.eventPools);
      this.logger.info(
        `starting to listen to new non-null pool: ${key}. Already following ${allEventPools
          // Not that I like this reduce, but since it is done only on initialization, expect this to be ok
          .reduce(
            (acc, curr) => (curr !== null ? ++acc : acc),
            0,
          )} non-null pools or ${allEventPools.length} total pools`,
      );
    }
```

KanievskyiDanylo (approved):

PR #573, opened 2023-11-13
BACK-1380: Restrict Dexalot on pool level

JackieJoo:


src/dex/dexalot/dexalot.ts
```diff
@@ -579,6 +576,7 @@ export class Dexalot extends SimpleExchange implements IDex<DexalotData> {
       const makerToken = normalizedDestToken;
       const takerToken = normalizedSrcToken;
 
+      const slippageBps = side === SwapSide.SELL ? BigNumber(1).minus(options.slippageFactor).multipliedBy(10000).toFixed(0) : options.slippageFactor.minus(1).multipliedBy(10000).toFixed(0);
```
aburkut, line 579: @JackieJoo I see that condition `side === SwapSide.SELL` is frequently used within preProcessTransaction method. Maybe it's worth to introduce isSell var `const isSell = side === SwapSide.SELL` ?

aburkut (approved):

src/dex/dexalot/dexalot.ts
JackieJoo: fixed

KanievskyiDanylo (approved):

PR #575, opened 2023-11-14
BACK-1378: Restrict SwaapV2 on pool level

JackieJoo:


src/dex/swaap-v2/swaap-v2.ts
```diff
@@ -23,6 +29,8 @@ import {
   SwaapV2APIParameters,
   SwaapV2QuoteError,
   TokensMap,
+  SwaapV2NotificationRequest,
```
KanievskyiDanylo, line 32: Unused variable

KanievskyiDanylo (approved):

src/dex/swaap-v2/swaap-v2.ts
JackieJoo: fixed

src/dex/swaap-v2/swaap-v2.ts
```diff
@@ -578,33 +593,62 @@ export class SwaapV2 extends SimpleExchange implements IDex<SwaapV2Data> {
...
+    await this.dexHelper.cache.hset(
+      this.runtimeMMsRestrictHashMapKey,
+      poolIdentifier,
+      Date.now().toString(),
     );
 
-    return result === 'true';
+    this.rateFetcher
+      .notify(SWAAP_BANNED_CODE, message, this.getNotifyReqParams())
+      .then((answer: SwaapV2NotificationResponse) => {
```
aburkut, line 627: @JackieJoo Could we use async/await ?

JackieJoo: we'd need to wrap it in a separate try-catch, seems easier to just use `.then().catch()`

aburkut (approved):

PR #579, opened 2023-11-15
BACK-1359: Encode SimpleSwap path with 1 DEX involved with Executor01

aburkut:


src/dex/weth/weth.ts
```diff
@@ -181,11 +183,15 @@ export class Weth
...
-      );
+      const opType = WethFunctions.withdraw;
+      // const withdrawWethData = this.simpleSwapHelper.encodeFunctionData(
+      //   opType,
+      //   [wethToken],
+      // );
+
+      const withdrawWethData = this.erc20Interface.encodeFunctionData(opType, [
+        destAmount,
+      ]);
```
KanievskyiDanylo, line 194: Why do we need to withdraw specific amount? I assume that it could lead to leaving tokens on Augustus

PR #591, opened 2023-11-24
Feat/back 1392

KanievskyiDanylo:


JackieJoo (approved):

src/dex/algebra/algebra.ts
```diff
@@ -730,30 +815,46 @@ export class Algebra extends SimpleExchange implements IDex<AlgebraData> {
...
-            deadline: getLocalDeadlineAsFriendlyPlaceholder(),
-            amountOut: destAmount,
-            amountInMaximum: srcAmount,
-            path,
-          };
+    let swapFunction;
+    let swapParams;
+
+    if (data.feeOnTransfer) {
+      swapFunction = AlgebraFunctions.exactInputWithFeeToken;
```
mwamedacen, line 822: I think it's worth adding an invariant check here to asses that we have exactly 1 pool involved.

So something like 
```javascript
assert(data.path.length != 1, 'LOGIC ERROR: multihop not supported for feeOnTransfer token');
```

PR #608, opened 2023-12-23
Back 1424/fix camelot v2 pool discovery

mwamedacen:


src/dex/camelot/camelot.ts
```diff
@@ -205,8 +213,58 @@ export class Camelot
...
+    );
+  }
+
+  private _constructPoolKey(token0: string, token1: string) {
+    const [_token0, _token1] =
+      token0.toLowerCase() < token1.toLowerCase()
+        ? [token0, token1]
+        : [token1, token0];
+
+    const poolKey = `${token0}-${token1}`.toLowerCase();
```
aburkut, line 232: @mwamedacen '${_token0}-${_token1}'.toLowerCase()?

src/dex/camelot/camelot.ts
```diff
@@ -205,8 +213,58 @@ export class Camelot
...
+   * Once the pool is created, it gets immediately flagged
+   */
+  onPoolCreatedDeleteFromNonExistingSet: OnPoolCreatedCallback = async ({
+    token0,
+    token1,
+  }) => {
+    const logPrefix = '[onPoolCreatedDeleteFromNonExistingSet]';
+
+    try {
+      const [_token0, _token1] =
```
aburkut, line 248: @mwamedacen not clear why we need this here since we have the same in _constuctPoolKey already

src/dex/camelot/camelot.ts
```diff
@@ -205,8 +213,58 @@ export class Camelot
...
+
+      this.newlyCreatedPoolKeys.add(poolKey);
+
+      // delete entry locally to let local instance discover the pool
+      delete this.pairs[poolKey];
+
+      this.logger.info(`${logPrefix} discovered new pool ${poolKey}`);
+    } catch (e) {
+      this.logger.error(
+        `${logPrefix} LOGIC ERROR on ack new pool (token0=${token0},token1=${token1})`,
```
aburkut, line 263: @mwamedacen why we are not logging actual error here?

src/dex/camelot/camelot.ts
```diff
@@ -44,6 +44,10 @@ import { SimpleExchange } from '../simple-exchange';
 import { applyTransferFee } from '../../lib/token-transfer-fee';
 import * as CALLDATA_GAS_COST from '../../calldata-gas-cost';
 import { SolidlyData } from '../solidly/types';
+import {
+  OnPoolCreatedCallback,
+  UniswapV2Factory,
+} from '../uniswap-v2/uniswap-v2-factory';
```
aburkut, line 50: @mwamedacen Seems the issue is actual for univ2 too. Why not implement the fix for it too?

mwamedacen: Prioritization

src/dex/camelot/camelot.ts
mwamedacen: Thx I will

src/dex/camelot/camelot.ts
mwamedacen: I kept same convention as in the code

aburkut: @mwamedacen I meant it seems to me that you used incorrect vars. From what I see it should be `_token0` instead of `token0` and `_token1` instead of `token1`. In other case I don't understand why you need tokens sorting above.

mwamedacen: good catch !

src/dex/camelot/camelot.ts
mwamedacen: you're right, this is deadcode

JackieJoo (approved):

aburkut (approved):

src/dex/uniswap-v2/uniswap-v2-factory.ts
```diff
@@ -0,0 +1,71 @@
...
+    }
+
+    return {};
+  }
+
+  async handleNewPool(event: LogDescription) {
+    const token0 = event.args.token0.toLowerCase();
+    const token1 = event.args.token1.toLowerCase();
+
+    await this.onPoolCreated({ token0, token1 });
```
KanievskyiDanylo, line 69: I think it makes sense to pass `pair` ([pool address](https://github.com/paraswap/paraswap-dex-lib/blob/bee98063e59db6faa82a0debf5ad481d172be51a/src/abi/uniswap-v2/uniswap-v2-factory.json#L38)) param here as well. Even though it's not used for Camelot, it could be relevant  for some of the cases
image

src/dex/uniswap-v2/uniswap-v2-factory.ts
```diff
@@ -0,0 +1,71 @@
...
+
+  constructor(
+    readonly dexHelper: IDexHelper,
+    parentName: string,
+    protected readonly factoryAddress: Address,
+    logger: Logger,
+    protected readonly onPoolCreated: OnPoolCreatedCallback,
+    mapKey: string = '',
+  ) {
+    super(parentName, `${parentName} Factory`, dexHelper, logger, true, mapKey);
```
KanievskyiDanylo, line 40: As we don't store state for factory, it assume it should also work as: 
```ts
super(parentName, `${parentName} Factory`, dexHelper, logger)
```

KanievskyiDanylo (approved):

src/dex/uniswap-v2/uniswap-v2-factory.ts
mwamedacen: could come handy but only for logs

PR #618, opened 2024-01-15
Feat/back 1441 rfq dynamic taker

mwamedacen:
resolves BACK-1441

### DIFF
--- a/src/dex/generic-rfq/fetch-mm-api.ts
+++ b/src/dex/generic-rfq/fetch-mm-api.ts
@@ -31,6 +31,10 @@ import { validateAndCast } from '../../lib/validators';
 
 const network = 1;
 
+const AgustusAddress = {
+  1: '0xdef171fe48cf0115b1d80b88dc8eab59176fee57',
+};
+
 const getEnv = (envName: string): string => {
   if (!process.env[envName]) {
     throw new Error(`Missing ${envName}`);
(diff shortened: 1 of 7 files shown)


src/dex/generic-rfq/fetch-mm-api.ts
```diff
@@ -31,6 +31,10 @@ import { validateAndCast } from '../../lib/validators';
 
 const network = 1;
 
+const AgustusAddress = {
+  1: '0xdef171fe48cf0115b1d80b88dc8eab59176fee57',
```
mwamedacen, line 35: file used for testing purposes only

src/dex/generic-rfq/rate-fetcher.ts
```diff
@@ -425,6 +427,7 @@ export class RateFetcher {
         this.dexHelper.config.data.network,
```
KanievskyiDanylo, line 427: I think it makes sense to add `takerAddress: addressSchema.required(),` to `firmRateResponseValidator.order` validator

src/dex/generic-rfq/fetch-mm-api.ts
KanievskyiDanylo: Typo issue, not a big deal though

src/dex/generic-rfq/rate-fetcher.ts
mwamedacen: Actually it was already there but I missed [this validation rule](https://github.com/paraswap/paraswap-dex-lib/blob/master/src/dex/generic-rfq/validators.ts#L171) which enforced augustus addresss. For now it's ok but we'll have to relax it for the future, doing it now

PR #626, opened 2024-03-05
Add E3 Dex on Arbitrum, Base and Fantom chain

ftm1337:
* E3 is the oldest fork of TraderJoe v2.1, present on the Fantom, Arbitrum One and the Base Network.
* It inherits all the attributes, interfaces and functions of TJv21
* Additionally, it expands it to make it compatible with Solidly-style revenue sharing,
* As well as a slightly altered core which makes it a little bit more gas & capital efficient.
* Such added functions do not impact how Swaps or Liquidity are made externally.
* Thus, it can directly re-use existing TraderJoe v2.1 adapters here.
* The Router input tuple param "VERSION" remains the same `2` as is with TJ v2.1.
* For backwards compatibility, E3 has placeholder forks of TJv2 and TJv1 in place on all 3 chains.

Further Docs: https://eliteness.network/docs/E3

Daily Volume: $1 Million

Age: 1 year


ftm1337:
Closing due to no response here

PR #636, opened 2024-04-04
BACK-1503: Re-assess need for poolconfig in uniswapV3 & forks

aburkut:


src/stateful-event-subscriber.ts
```diff
@@ -173,6 +181,10 @@ export abstract class StatefulEventSubscriber<State>
     this.isInitialized = true;
   }
 
+  protected getValueToPushIntoNewPoolsChannel() {
+    return {};
```
mwamedacen, line 185: please make it abstract or throw an error

src/constants.ts
```diff
@@ -9,6 +9,8 @@ export const NULL_ADDRESS = '0x0000000000000000000000000000000000000000';
 
 export const CACHE_PREFIX = 'dl';
 
+export const INIT_SERVICE_CACHE_PREFIX = 'is';
```
mwamedacen, line 12: let's stick to dl

aburkut: updated

src/stateful-event-subscriber.ts
aburkut: @mwamedacen I didn't get your point here. Why we need to make it abstract? Now getValueToPushIntoNewPoolsChannel method contains default implementation for all pools.

aburkut: updated

mwamedacen (approved):

PR #643, opened 2024-04-09
BACK-1506: Hashflow fix mm bug, add logs

JackieJoo:


src/dex/hashflow/hashflow.ts
```diff
@@ -714,7 +714,7 @@ export class Hashflow extends SimpleExchange implements IDex<HashflowData> {
     } catch (e) {
       if (
         e instanceof Error &&
-        e.message.endsWith('User is restricted from using Hashflow')
+        e.message.includes('User is restricted from using Hashflow')
```
mwamedacen, line 717: I suggest to lower case message and check just for "user is restricted"

src/dex/hashflow/hashflow.ts
```diff
@@ -726,6 +726,9 @@ export class Hashflow extends SimpleExchange implements IDex<HashflowData> {
             `${this.dexKey}-${this.network}: Market Maker ${mm} failed to build transaction on side ${side} with too strict slippage. Skipping restriction`,
           );
         } else {
+          this.logger.error(
```
mwamedacen, line 729: let's keep it warn (same as most of the errors and [this log](https://github.com/paraswap/paraswap-dex-lib/blob/0b688cd16727616b16c26130adc5d19a15f07792/src/dex/hashflow/hashflow.ts#L741)).

src/dex/hashflow/hashflow.ts
JackieJoo: done

src/dex/hashflow/hashflow.ts
JackieJoo: done

mwamedacen (approved):

PR #646, opened 2024-04-12
BACK-1512: improve hashflow mm restriction

JackieJoo:


src/dex/hashflow/constants.ts
```diff
@@ -20,3 +20,37 @@ export const HASHFLOW_GAS_COST = 100_000;
 
 export const HASHFLOW_MIN_SLIPPAGE_FACTOR_THRESHOLD_FOR_RESTRICTION =
   new BigNumber('0.001');
+
+export const RESTRICT_79_TTL_MS = 60 * 10 * 1000; // 10 mins
```
aburkut, line 24: @JackieJoo I would you use a bit more reasonable naming here - for example RESTRICT_BELOW_MINIMUM_AMOUNT_TTL_MS

src/dex/hashflow/constants.ts
```diff
@@ -20,3 +20,37 @@ export const HASHFLOW_GAS_COST = 100_000;
...
+export const RESTRICT_SLIPPAGE_TTL_MS = 60 * 5 * 1000; // 5 mins
+export const RESTRICT_84_TTL_MS = 60 * 10 * 1000; // 10 mins
```
aburkut, line 26: RESTRICT_RATE_LIMIT_TTL_S

aburkut: Actual for the rest of the constants...

JackieJoo: I'd rather leave error codes in the ttl name, because we have a mapping `ERROR_CODE_TO_RESTRICT_TTL`, and it's easy to see what correspond where, and we use those codes as indicator that this error is known and comes from mm directly

JackieJoo: Also error without error codes are ours, for example `MISSING_QUOTE_DATA`, `MISSING_SIGNATURE_DATA`, they don't have codes in the name - `RESTRICT_MISSING_QUOTE_DATA_TTL_MS`, etc

KanievskyiDanylo (approved):

aburkut (approved):

KanievskyiDanylo (approved):

PR #658, opened 2024-04-24
BACK-1532: Integrate PharaohV1

JackieJoo:


src/dex/solidly/forks-override/pharaohV1.ts
```diff
@@ -0,0 +1,253 @@
...
+  reserve1: string;
+  token1: string;
+};
+
+export type PharaohV1SubgraphToken = {
+  id: string;
+  decimals: string;
+};
+
+export class PharaohV1 extends Solidly {
```
aburkut, line 38: @JackieJoo I think we don't need this separate implementation since it's a copy paste from [Ramses class](https://github.com/paraswap/paraswap-dex-lib/blob/feat/BACK-1532/src/dex/solidly/forks-override/ramses.ts#L38). Let's extend PharaohV1 from Ramses.

src/dex/solidly/config.ts
```diff
@@ -159,6 +159,18 @@ export const SolidlyConfig: DexConfigMap<DexParams> = {
       feeCode: 0,
     },
   },
+  PharaohV1: {
```
aburkut, line 162: @JackieJoo I would suggest to rename Pharaoh (which is UniV3 fork) to PharaohV2 to have consistency with Ramses.

src/dex/solidly/forks-override/pharaohV1.ts
JackieJoo: done, thx

mwamedacen (approved):

aburkut (approved):

PR #674, opened 2024-05-10
Update Integral integration to use event driven update model

IntegralHQ:


IntegralHQ:
Hi All,

A few notes with this PR:

1. We made some changes to Oracle.ts in in Uniswap-v3/contract-math so that we could reuse their oracle code logic. Our system relies on the TWAP oracle built into Uniswap. Since they already had implemented the majority of the code (with some inadvertent bugs), we simply fixed the bugs and used the existing libraries.

Since there could be some concern that these changes may impact the Uniswap integration, I've added comments to the PR with explanations of our changes.

2. Behind the scenes of the TWAP oracle from Uniswap is the accumulation of ticks. The number of ticks can be quite large depending the cardinality set on each Uniswap pair (it's user configurable). Since there is quite of data to pull initially, tests run a bit longer. Once the data has been pulled, the internal state is kept in sync by listening to events.

src/dex/uniswap-v3/contract-math/Oracle.ts
```diff
@@ -23,7 +23,7 @@ export class Oracle {
   ): OracleObservation {
     const delta = blockTimestamp - last.blockTimestamp;
     return {
-      blockTimestamp: state.blockTimestamp,
+      blockTimestamp,
```
IntegralHQ, line 26: Here we simply matched the .ts implementation with the .sol implementation: https://github.com/Uniswap/v3-core/blob/main/contracts/libraries/Oracle.sol#L39

You can see that new observations also update the blockTimestamp in the Solidity code.

Furthermore, since Uniswap's pricing does not depend on the oracle logic, you can be rest assured this won't negatively impact the Uniswap quoting mechanics.

src/dex/uniswap-v3/contract-math/Oracle.ts
```diff
@@ -93,12 +93,12 @@ export class Oracle {
     let beforeOrAt;
     let atOrAfter;
     while (true) {
-      i = (l + r) / 2;
+      i = Math.floor((l + r) / 2);
```
IntegralHQ, line 96: This is another bug fix to the original Uniswap .ts implementation. In Solidity, division is done on integers so there's implicitly a floor. We added floor manually to match the EVM execution: https://github.com/Uniswap/v3-core/blob/main/contracts/libraries/Oracle.sol#L164

src/dex/uniswap-v3/contract-math/Oracle.ts
```diff
@@ -93,12 +93,12 @@ export class Oracle {
...
 
       beforeOrAt = state.observations[i % cardinality];
 
       // we've landed on an uninitialized tick, keep searching higher (more recently)
-      if (!beforeOrAt.initialized) {
+      if (!beforeOrAt || !beforeOrAt.initialized) {
```
IntegralHQ, line 101: In Solidity, accessing a member of an undefined struct won't cause any issues since the values will be returned as 0 or empty. But in Javascript, accessing a memory on a undefined object will cause the system to crash. We simply added a check that the object should be defined before accessing its member here: https://github.com/Uniswap/v3-core/blob/main/contracts/libraries/Oracle.sol#L169

src/dex/uniswap-v3/contract-math/Oracle.ts
```diff
@@ -150,7 +150,8 @@ export class Oracle {
     }
 
     beforeOrAt = state.observations[(index + 1) % cardinality];
-    if (!beforeOrAt.initialized) beforeOrAt = state.observations[0];
+    if (!beforeOrAt || !beforeOrAt.initialized)
```
IntegralHQ, line 153: The change here is the same as the one in 101: https://github.com/Uniswap/v3-core/blob/main/contracts/libraries/Oracle.sol#L223

PR #677, opened 2024-05-14
feat(BACK-1551): add rswETH pool

nzrii-kysel:


src/dex/swell/rsweth-pool.ts
Wertusser, line 1: It looks like this file is identical to `sweth-pool.ts` and can be removed, or am I missing something?

Wertusser (approved):

src/dex/swell/rsweth-pool.ts
nzrii-kysel: state and event interfaces are slightly different

KanievskyiDanylo (approved):

aburkut (approved):

PR #685, opened 2024-05-29
BACK-1542: Integrate Idle

JackieJoo:


src/dex/idle-dao/idle-dao.ts
```diff
@@ -0,0 +1,372 @@
...
+
+      returnAmountPos = extractReturnAmountPosition(
+        this.cdo,
+        PoolFunctions[`deposit${data.idleToken.tokenType}`],
+      );
+      return [
+        this.cdo,
+        data.idleToken.cdoAddress,
+        PoolFunctions[`deposit${data.idleToken.tokenType}`],
+        [srcAmount, recipient],
```
nzrii-kysel, line 333: why are we passing recipient as a referral? and why are we using deposit with referral in the first place?

src/dex/idle-dao/scripts.ts
nzrii-kysel, line 1: i think this can be deleted

src/dex/idle-dao/config.ts
```diff
@@ -0,0 +1,309 @@
+import { DexParams, IdleToken } from './types';
+import { DexConfigMap, AdapterMappings } from '../../types';
+import { Network, SwapSide } from '../../constants';
+
+export const Config: DexConfigMap<DexParams> = {
+  IdleDao: {
+    [Network.MAINNET]: {
+      lendingGasCost: 328 * 1000,
+      factoryAddress: '0x3c9916bb9498f637e2fa86c2028e26275dc9a631',
+    },
```
KanievskyiDanylo, line 10: What about other chains? (polygon, zkevm, optimism)

src/dex/idle-dao/config.ts
```diff
@@ -0,0 +1,309 @@
...
+};
+
+export const endpoints: Record<number, string> = {
+  [Network.MAINNET]: 'https://api.idle.finance/pools',
+  [Network.POLYGON]: 'https://api-polygon.idle.finance/pools',
+  [Network.OPTIMISM]: 'https://api-optimism.idle.finance/pools',
+  [Network.ZKEVM]: 'https://api-zkevm.idle.finance/pools',
+};
+
+export const TOKEN_LISTS: IdleToken[] = [
```
KanievskyiDanylo, line 28: Assume that this token list could be extended for other chains, maybe then add something like `export const TOKEN_LISTS: Record<network, IdleToken[]> = {}`?

KanievskyiDanylo: Also, does it make sense to move it separate json file?

src/dex/idle-dao/tokens.ts
```diff
@@ -0,0 +1,90 @@
+import { Token } from '../../types';
+import { Network } from '../../constants';
+
+import { IdleToken } from './types';
+import { IDexHelper } from '../../dex-helper';
+import CDO_ABI from '../../abi/idle-dao/idle-cdo.json';
+import { AbiItem } from 'web3-utils';
+
+const poolsByTokenAddress: { [address: string]: IdleToken[] } = {};
```
KanievskyiDanylo, line 9: Should not it be also split by chain? like `const poolsByTokenAddress: {[network: number]: {[address: string]: IdleToken[]} } = {};`

src/dex/idle-dao/utils.ts
```diff
@@ -0,0 +1,215 @@
...
+    results
+      .filter(result => !!result.tokenType)
+      .map(result => result.idleAddress),
+    erc20Interface,
+    multiWrapper,
+  );
+
+  const idleTokenDecimals = await _getIdleTokenDecimals(
+    blockNumber,
+    results.filter(result => !result.tokenType).map(result => result.address),
```
KanievskyiDanylo, line 192: Incorrect filter? assume it should be `results.filter(result => !!result.tokenType).map(result => result.address)`

src/dex/idle-dao/idle-dao.ts
```diff
@@ -0,0 +1,372 @@
...
+  }
+
+  async getTopPoolsForToken(
+    tokenAddress: Address,
+    limit: number,
+  ): Promise<PoolLiquidity[]> {
+    if (!this.tokenList) {
+      const blockNumber =
+        await this.dexHelper.web3Provider.eth.getBlockNumber();
+      this.initializePricing(blockNumber);
```
KanievskyiDanylo, line 358: can do the same as with Stata to make `getTopPoolsForToken` independant from pricing logic

src/dex/idle-dao/idle-dao.ts
```diff
@@ -0,0 +1,372 @@
...
+    }
+    const idleTokens: IdleToken[] = getPoolsByTokenAddress(tokenAddress);
+
+    return idleTokens
```
KanievskyiDanylo, line 362: I am not very aware of the current dex, but want to clarify if it makes sense to `idleAddress -> address` connectors?
Because as I can see now, it only returns new connectors for the usual tokens (`address` field in the list), but maybe it's worth to know that `address` is the connector for `idleAddress` as well? (for `idleAddress` return `address` connector)

src/dex/idle-dao/idle-dao.ts
```diff
@@ -0,0 +1,372 @@
...
+      this.network,
+      TOKEN_LIST_CACHE_KEY,
+      TOKEN_LIST_TTL_SECONDS,
+      JSON.stringify(this.tokenList),
+    );
+
+    setTokensOnNetwork(this.network, this.tokenList, this.dexHelper);
+  }
+
+  getAdapters(side: SwapSide): { name: string; index: number }[] | null {
```
KanievskyiDanylo, line 135: better to return null before v5 support

src/dex/idle-dao/idle-dao.ts
```diff
@@ -0,0 +1,372 @@
...
+      const blockNumber =
+        await this.dexHelper.web3Provider.eth.getBlockNumber();
+      this.initializePricing(blockNumber);
+    }
+    const idleTokens: IdleToken[] = getPoolsByTokenAddress(tokenAddress);
+
+    return idleTokens
+      .map((idleToken: IdleToken) => ({
+        // liquidity is infinite, tokens are minted when swapping for idle tokens
+        liquidityUSD: 1e12,
```
KanievskyiDanylo, line 365: $1m could give too much priority for this token, especially for L2 chains. maybe 100_000 would be better?

src/dex/idle-dao/idle-dao.ts
JackieJoo: it's been done by the dex team, don't see any issue with that, maybe want to track swaps

src/dex/idle-dao/utils.ts
JackieJoo: no, it's correct there, tokens that don't have tokenType would have `address`
tokens with `tokenType`, would have `idleAddress` and `cdoAddress`

src/dex/idle-dao/scripts.ts
JackieJoo: done

src/dex/idle-dao/config.ts
JackieJoo: done

src/dex/idle-dao/tokens.ts
JackieJoo: done

src/dex/idle-dao/idle-dao.ts
JackieJoo: done

src/dex/idle-dao/idle-dao.ts
JackieJoo: done

src/dex/idle-dao/idle-dao.ts
```diff
@@ -0,0 +1,372 @@
...
+    setTokensOnNetwork(this.network, this.tokenList, this.dexHelper);
+  }
+
+  getAdapters(side: SwapSide): { name: string; index: number }[] | null {
+    return this.adapters[side] ? this.adapters[side] : null;
+  }
+
+  private _getPoolIdentifier(srcToken: Token, destToken: Token): string {
+    return (
+      this.dexKey +
```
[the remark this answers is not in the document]
JackieJoo: done

src/dex/idle-dao/idle-dao.ts
JackieJoo: reduced

src/dex/idle-dao/idle-dao.ts
JackieJoo: did additional logic to check for idle connectors

src/dex/idle-dao/idle-dao.ts
nzrii-kysel: the only issue is that it is a bit more expensive, but its fine if the dex team has reasons for using it

0xNazarii (approved):

PR #700, opened 2024-06-10
BACK-1571: Integrate Stata

JackieJoo:


src/dex/aave-v3-stata/aave-v3-stata.ts
```diff
@@ -0,0 +1,482 @@
...
+      ) {
+        return null;
+      }
+    } else {
+      if (otherAddressLower !== stataToken.underlying.toLowerCase()) {
+        return null;
+      }
+    }
+
+    const stata = isSrcStata ? srcToken : destToken;
```
nzrii-kysel, line 191: this var seems redundant, you can get lowercase address from `stataToken` above

src/dex/aave-v3-stata/aave-v3-stata.ts
```diff
@@ -0,0 +1,482 @@
...
+          blockNumber,
+        );
+        this.state[stataAddressLower] = {
+          blockNumber,
+          rate: results[0].returnData,
+        };
+        this.dexHelper.cache.setex(
+          this.dexKey,
+          this.network,
+          'state',
```
nzrii-kysel, line 225: missing stata address in cache key

src/abi/aavev3stata/Factory.json
```diff
@@ -0,0 +1,128 @@
+[
```
KanievskyiDanylo, line 1: Unused file

src/dex/aave-v3-stata/tokens.ts
```diff
@@ -0,0 +1,48 @@
...
+  if (token.underlying === addressLower) return TokenType.UNDERLYING;
+  if (token.underlyingAToken === addressLower) return TokenType.A_TOKEN;
+  return TokenType.UNKNOWN;
+}
+
+export function getTokenFromAddress(
+  network: Network,
+  address: string,
+): StataToken {
+  return TokensByAddress[network][address.toLowerCase()];
```
KanievskyiDanylo, line 47: `TokensByAddress[network]` might be empty

src/abi/aavev3stata/Factory.json
JackieJoo: It's used [here](https://github.com/paraswap/paraswap-dex-lib/pull/700/files#diff-5d7c647f1faef6ba75b678ad3eb6d13839904f90b91e91e53f0b94a07567d8abR73-R76)

src/dex/aave-v3-stata/aave-v3-stata.ts
```diff
@@ -0,0 +1,482 @@
...
+        blockNumber: 0,
+        rate: 0n,
+      };
+    });
+  }
+
+  // Returns the list of contract adapters (name and index)
+  // for a buy/sell. Return null if there are no adapters.
+  getAdapters(side: SwapSide): { name: string; index: number }[] | null {
+    return this.adapters[side] ? this.adapters[side] : null;
```
KanievskyiDanylo, line 116: Better to get rid of `adapters`, as we won't support it on V5 (return `null` here)

JackieJoo: We plan to add support for V5 also for this integration, so let them be there

src/dex/aave-v3-stata/tokens.ts
JackieJoo: fixed

src/dex/aave-v3-stata/aave-v3-stata.ts
```diff
@@ -0,0 +1,482 @@
...
+        ]);
+      }
+    }
+
+    return {
+      needWrapNative: this.needWrapNative,
+      dexFuncHasRecipient: true,
+      exchangeData: swapData,
+      targetExchange: exchange,
+      returnAmountPos: undefined,
```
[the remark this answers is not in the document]
JackieJoo: fixed

src/dex/aave-v3-stata/aave-v3-stata.ts
JackieJoo: fixed

src/dex/aave-v3-stata/aave-v3-stata.ts
JackieJoo: fixed

src/dex/aave-v3-stata/aave-v3-stata.ts
```diff
@@ -0,0 +1,492 @@
...
+    super(dexHelper, dexKey);
+    this.logger = dexHelper.getLogger(dexKey);
+  }
+
+  // Initialize pricing is called once in the start of
+  // pricing service. It is intended to setup the integration
+  // for pricing requests. It is optional for a DEX to
+  // implement this function
+  async initializePricing(blockNumber: number) {
+    let cachedTokenList = await this.dexHelper.cache.getAndCacheLocally(
```
KanievskyiDanylo, line 76: no need to cache locally

JackieJoo: why? it's gonna be both local/distant

src/dex/aave-v3-stata/aave-v3-stata.ts
```diff
@@ -0,0 +1,492 @@
...
+
+    setTokensOnNetwork(this.network, tokenList);
+
+    // init state for all tokens as empty
+    tokenList.forEach(token => {
+      this.state[token.address] = {
+        blockNumber: 0,
+        rate: 0n,
+      };
+    });
```
KanievskyiDanylo, line 110: not sure why this initialization is done only when tokens weren't cached. I think it's worth to do for the cached tokens as well:
```ts
    if (cachedTokenList !== null) {
      setTokensOnNetwork(this.network, JSON.parse(cachedTokenList));
      tokenList.forEach(token => {
      this.state[token.address] = {
        blockNumber: 0,
        rate: 0n,
      };
    });

    return;
    }
```

src/dex/aave-v3-stata/aave-v3-stata.ts
KanievskyiDanylo: let's still return null here, to prevent any erros on V5 build

src/dex/aave-v3-stata/aave-v3-stata.ts
```diff
@@ -0,0 +1,492 @@
...
+
+  // Returns the list of contract adapters (name and index)
+  // for a buy/sell. Return null if there are no adapters.
+  getAdapters(side: SwapSide): { name: string; index: number }[] | null {
+    return this.adapters[side] ? this.adapters[side] : null;
+  }
+
+  private _getPoolIdentifier(srcToken: Token, destToken: Token): string {
+    return (
+      this.dexKey +
```
KanievskyiDanylo, line 121: please add `-` after `this.dexKey` to ease debugging in logs for this dex

src/dex/aave-v3-stata/aave-v3-stata.ts
JackieJoo: you're right, fixed

src/dex/aave-v3-stata/aave-v3-stata.ts
JackieJoo: done

src/dex/aave-v3-stata/aave-v3-stata.ts
JackieJoo: done

src/dex/aave-v3-stata/aave-v3-stata.ts
```diff
@@ -0,0 +1,507 @@
...
+              return this.previewWithdraw(amount, rate);
+            } else {
+              return this.previewMint(amount, rate);
+            }
+          }
+        }),
+        unit: getBigIntPow(
+          (side === SwapSide.SELL ? destToken : srcToken).decimals,
+        ),
+        gasCost: 400_000, // 250_000 from underlying, far less from aToken
```
KanievskyiDanylo, line 266: Redundantly high gasCost could deprioritize pricing against other dexes, better to leave todo and update after tests

src/dex/aave-v3-stata/aave-v3-stata.ts
```diff
@@ -0,0 +1,507 @@
...
+    // TODO: complete me!
+  }
+
+  // Returns list of top pools based on liquidity. Max
+  // limit number pools should be returned.
+  async getTopPoolsForToken(
+    tokenAddress: Address,
+    limit: number,
+  ): Promise<PoolLiquidity[]> {
+    //TODO: complete me!
```
KanievskyiDanylo, line 450: Would be nice to know if there's need for such connectors. I assume we might miss some routes if new tokens (stata) will be available on other dexes like Uniswap

src/dex/aave-v3-stata/aave-v3-stata.ts
KanievskyiDanylo: because `getAndCacheLocally` with current params is called only once, therefore no need to store it locally

src/dex/aave-v3-stata/aave-v3-stata.ts
```diff
@@ -0,0 +1,507 @@
...
+        swapData = AaveV3Stata.stata.encodeFunctionData(StataFunctions.redeem, [
+          srcAmount,
+          recipient, // receiver
+          executorAddress, // owner
+          destType === TokenType.UNDERLYING, // withdraw from aToken
+        ]);
+
+        returnAmountPos = extractReturnAmountPosition(
+          AaveV3Stata.stata,
+          StataFunctions.redeem,
```
KanievskyiDanylo, line 386: You should pass function name, not function interface (`redeem(uint256,address,address,bool)`)

src/dex/aave-v3-stata/aave-v3-stata.ts
```diff
@@ -0,0 +1,507 @@
...
+          '',
+          1,
```
KanievskyiDanylo, line 388: Internal type is `uint256` not `uint256[]`, so no need to extract dynamic params

src/dex/aave-v3-stata/aave-v3-stata.ts
KanievskyiDanylo: Though it might work
image

src/dex/aave-v3-stata/aave-v3-stata.ts
JackieJoo: updated

src/dex/aave-v3-stata/aave-v3-stata.ts
JackieJoo: There are 2 functions with the same name for both deposit & redeem, we're using the ones with 4 input params

src/dex/aave-v3-stata/aave-v3-stata.ts
JackieJoo: We get 2 values back [here](https://github.com/paraswap/paraswap-dex-lib/pull/700/files#diff-1deec285bd07ad57d19bd1f7a4a229997802c9b4c2fddd2d901412fb32b1e5b3R659-R661)

KanievskyiDanylo: This param works if returned type is array. For now this function returns tuple. Then we should adapt function to handle tuples, not just arrays

KanievskyiDanylo: Ok, as only a tuple is returned, then you can use `RETURN_AMOUNT_POS_32` instead of interacting with `extractReturnAmountPosition`

If you need to get first element, not second, from the tuple, then you should use `RETURN_AMOUNT_POS_0`

JackieJoo: I'll leave it undefined for this case, and leave a TODO to improve, would need to retest thoroughly

KanievskyiDanylo: No need to postpone this, it's an easy update.

Just use return_amount_pos_32 as const value if you want to extract second uint256 value from static calldata.

calladata encoding [link](https://medium.com/@scourgedev/deep-dive-into-abi-encode-types-padding-and-disassembly-84472f1b4543)

src/dex/aave-v3-stata/aave-v3-stata.ts
JackieJoo: done

src/dex/aave-v3-stata/aave-v3-stata.ts
JackieJoo: done

nzrii-kysel (approved):

src/dex/aave-v3-stata/aave-v3-stata.ts
sakulstra: how exactly does this method work?

Currently most liquidity is in balancer v2 e-cpl rehype pools. Anyhow would be good to have this merged soonish to have basic support and improve afterwards. Happy to help with that.

KanievskyiDanylo (approved):

PR #713, opened 2024-06-20
BACK-1596: Add fix for cached price levels for SwaapV2

aburkut:


src/dex/swaap-v2/rate-fetcher.ts
```diff
@@ -119,27 +119,30 @@ export class RateFetcher {
...
-        }
-        const levels = resp.levels[pairName];
+    const levels = Object.keys(resp.levels).reduce<
+      Record<string, SwaapV2PriceLevels>
+    >((memo, pairName) => {
+      const pair = resp.levels[pairName];
+      if (!pair) {
+        return memo;
+      }
+      const levels = resp.levels[pairName];
```
nzrii-kysel, line 129: `pairs` and `levels` are the same

src/dex-helper/dummy-dex-helper.ts
```diff
@@ -146,6 +146,10 @@ class DummyCache implements ICache {
   }
 
   async hget(mapKey: string, key: string): Promise<string | null> {
+    if (!this.hashStorage[mapKey]) {
+      return null;
+    }
+
     return this.hashStorage[mapKey][key];
```
nzrii-kysel, line 153: this would ensure we never return `undefined`, wdyt?
```suggestion
    return this.hashStorage[mapKey]?.[key] ?? null;
```

src/dex/swaap-v2/rate-fetcher.ts
aburkut: @nzrii-kysel fixed, thx

src/dex-helper/dummy-dex-helper.ts
aburkut: @nzrii-kysel  fixed

nzrii-kysel (approved):

PR #718, opened 2024-06-28
fix: remove balancer pools thats in recovery mode

Wertusser:


src/dex/balancer-v2/balancer-v2.ts
```diff
@@ -415,16 +416,16 @@ export class BalancerV2EventPool extends StatefulEventSubscriber<PoolStateMap> {
       throw new Error('Unable to fetch pools from the subgraph');
 
     const poolsMap = keyBy(data.pools, 'address');
-    const allPools: SubgraphPoolBase[] = data.pools.map(
-      (pool: Omit<SubgraphPoolBase, 'mainTokens'>) => ({
+    const allPools: SubgraphPoolBase[] = data.pools
+      .filter((pool: SubgraphPoolBase) => !pool.isInRecoveryMode)
```
Wertusser, line 420: Its better to add this filter in subgraph query, but for most of pools `isInRecoveryMode` is empty

KanievskyiDanylo: better still to have this in where query to cover all cases. no need to fetch pools that are not used

Wertusser: fixed

aburkut (approved):

PR #737, opened 2024-07-15
Add CurveV1Factory and CurveV1StableNg multihop support for v5

aburkut:


Wertusser (approved):

src/dex/curve-v1-factory/curve-v1-factory.ts
```diff
@@ -1115,11 +1141,21 @@ export class CurveV1Factory
       hexZeroPad(hexlify(blockNumber), 16),
     ]);
 
+    let wrapFlag = 0;
+
+    if (this.needWrapNative && isETHAddress(srcToken)) {
+      wrapFlag = 1; // wrap src eth
+    } else if (!this.needWrapNative && isETHAddress(srcToken)) {
+      wrapFlag = 3; // send eth
```
KanievskyiDanylo, line 1149: better to specify that this flag just add msg.value to router call

src/dex/curve-v1-factory/curve-v1-factory.ts
```diff
@@ -1187,14 +1216,70 @@ export class CurveV1Factory
...
+      item.j,
+      item.underlyingSwap
+        ? CurveRouterSwapType.exchange_underlying
+        : CurveRouterSwapType.exchange,
+      CurveRouterPoolType.stable,
+      item.n_coins,
+    ]);
+
+    while (swapParams.length < swapParamsLength) {
+      swapParams.push([0, 0, 0, CurveRouterPoolType.non_stable, 0]);
```
KanievskyiDanylo, line 1262: What if StableNgPools are used in multhihop, will it work?

src/dex/curve-v1-factory/curve-v1-factory.ts
```diff
@@ -1187,14 +1216,70 @@ export class CurveV1Factory
...
+      path.push(NULL_ADDRESS);
+    }
+
+    const swapParams: CurveV1RouterSwapParam[] = data.path.map(item => [
+      item.i,
+      item.j,
+      item.underlyingSwap
+        ? CurveRouterSwapType.exchange_underlying
+        : CurveRouterSwapType.exchange,
+      CurveRouterPoolType.stable,
```
KanievskyiDanylo, line 1257: Same here, if it will work for non stable pools

src/dex/curve-v1-factory/curve-v1-factory.ts
aburkut: updated

src/dex/curve-v1-factory/curve-v1-factory.ts
aburkut: Depending on this param CurveRouter decided which interface to assign for a pool in price route. In our case CurveRouterPoolType.stable will work for CurveV1StableNg and CurveV1Factory

src/dex/curve-v1-factory/curve-v1-factory.ts
aburkut: better to pass 0 instead of CurveRouterPoolType.non_stable to avoid confusion

KanievskyiDanylo (approved):

src/dex/curve-v1-factory/curve-v1-factory.ts
aburkut: updated

PR #755, opened 2024-07-30
BACK-1656: Integrate Infusion

JackieJoo:


src/dex/infusion/infusion.ts
```diff
@@ -0,0 +1,695 @@
...
+    routerMethod = isETHAddress(dest)
+      ? InfusionRouterFunctions.sellExactToken
+      : routerMethod;
+
+    routerArgs = isETHAddress(src)
+      ? [
+          destAmount,
+          [{ from, to, stable }],
+          recipient,
+          Math.floor(new Date().getTime()) + 120,
```
Wertusser, line 652: `getTime()` returns timestamp in milliseconds, Probably lost the division by 1000
```suggestion
          Math.floor(new Date().getTime() / 1000) + 120,
```

src/dex/infusion/infusion.ts
```diff
@@ -0,0 +1,695 @@
...
+    this.pairs[key] = pair;
+    return pair;
+  }
+
+  async batchCatchUpPairs(pairs: [Token, Token][], blockNumber: number) {
+    if (!blockNumber) return;
+    const pairsToFetch: InfusionPair[] = [];
+    for (const _pair of pairs) {
+      for (const stable of [false, true]) {
+        const pair = await this.findInfusionPair(_pair[0], _pair[1], stable);
```
Wertusser, line 170: Could be improved with `await Promise.all([...])`

src/dex/infusion/infusion.ts
```diff
@@ -0,0 +1,695 @@
...
+    destAmount: NumberAsString,
+    recipient: Address,
+    data: InfusionData,
+    side: SwapSide,
+  ): DexExchangeParam {
+    if (side === SwapSide.BUY) throw new Error(`Buy not supported`);
+
+    let routerMethod: any;
+    let routerArgs: InfusionParam;
+    const stable = isStablePair(this.network, src, dest);
```
Wertusser, line 631: Can we somehow add an `isStable` parameter to `InfusionData` so that we don't need to use a this util function with hardcoded list of stables?

src/dex/infusion/config.ts
```diff
@@ -0,0 +1,27 @@
...
+      // updatable fees on the factory without event
+      stableFee: 2,
+      volatileFee: 2,
+      poolGasCost: 180 * 1000,
+      feeCode: 2,
+    },
+  },
+};
+
+export const Adapters: Record<number, AdapterMappings> = {
```
aburkut, line 23: @JackieJoo Let's remove adapters since the integration will be on v6 only and these adapters are not actually used.

src/dex/infusion/infusion.ts
```diff
@@ -0,0 +1,695 @@
...
+
+    this.router =
+      routerAddress !== undefined
+        ? routerAddress
+        : InfusionConfig[dexKey][network].router || '';
+
+    this.feeFactor =
+      InfusionConfig[dexKey][network].feeFactor || this.feeFactor;
+
+    // this.exchangeRouterInterface = new Interface(infusionRouterABI);
```
aburkut, line 135: @JackieJoo Let's remove this commented line if it's not needed

src/dex/infusion/infusion.ts
```diff
@@ -0,0 +1,695 @@
...
+          this.feeFactor,
+        );
+  }
+
+  async getBuyPrice(
+    priceParams: InfusionPoolOrderedParams,
+    srcAmount: bigint,
+  ): Promise<bigint> {
+    if (priceParams.stable) throw new Error(`Buy not supported`);
+    return Uniswapv2ConstantProductPool.getBuyPrice(
```
aburkut, line 279: @JackieJoo I don't understand why we need to return something here if we don't support buy for this integration
BACK-1656: Integrate Infusion by JackieJoo · Pull Request #755 · paraswap:paraswap-dex-lib 🔊 2024-08-23 15-13-00

src/dex/infusion/infusion.ts
```diff
@@ -0,0 +1,695 @@
...
+  ): Promise<PoolLiquidity[]> {
+    if (!this.subgraphURL) return [];
+
+    let stableFieldKey = 'isStable';
+
+    // if (this.dexKey.toLowerCase() === 'infusion') {
+    //   stableFieldKey = 'stable';
+    // } else if (this.dexKey.toLowerCase() !== 'infusionv2') {
+    //   stableFieldKey = 'isStable';
+    // }
```
aburkut, line 443: @JackieJoo Do we need to keep these commented lines?

src/dex/infusion/infusion.ts
```diff
@@ -0,0 +1,695 @@
...
+import _ from 'lodash';
+import { NumberAsString, SwapSide } from '@paraswap/core';
+import { Interface, AbiCoder } from '@ethersproject/abi';
+import { InfusionStablePool } from './infusion-stable-pool';
+import { Uniswapv2ConstantProductPool } from '../uniswap-v2/uniswap-v2-constant-product-pool';
+import {
+  PoolState,
+  InfusionData,
+  InfusionPair,
+  InfusionPool,
```
aburkut, line 35: @JackieJoo unused var

src/dex/infusion/infusion.ts
```diff
@@ -0,0 +1,695 @@
...
+import { applyTransferFee } from '../../lib/token-transfer-fee';
+import { isStablePair } from './utils/isStablePair';
+
+export enum InfusionRouterFunctions {
+  sellExactEth = 'swapExactETHForTokens',
+  sellExactToken = 'swapExactTokensForETH',
+  swapExactIn = 'swapExactTokensForTokens',
+}
+
+export interface SmardexData extends Omit<UniswapV2Data, 'feeFactor'> {
```
aburkut, line 49: @JackieJoo unused interface

src/dex/infusion/infusion.ts
JackieJoo: But it's how we basically signify that a dex doesn't support BUY, you just return null for BUY side pricing, not sure what's the problem?

JackieJoo: ah, got the point, checking

src/dex/infusion/infusion.ts
JackieJoo: In any case we'd need to call this function, as I can see there's finite number of tokens considered as stables, so seems ok to leave as is

src/dex/infusion/infusion.ts
JackieJoo: done

src/dex/infusion/infusion.ts
JackieJoo: done

src/dex/infusion/config.ts
JackieJoo: done

src/dex/infusion/infusion.ts
JackieJoo: done

src/dex/infusion/infusion.ts
JackieJoo: fixed

src/dex/infusion/infusion.ts
JackieJoo: fixed

src/dex/infusion/infusion.ts
JackieJoo: fixed

src/dex/infusion/infusion.ts
JackieJoo: fixed

src/dex/infusion/infusion.ts
Wertusser: Not quite sure why then `InfusionData` was updated. There may be other pools with other “stable” tokens that may not be on the list, or pools that have tokens from this hard-coded list but this pool is not "stable"

JackieJoo: as I can see it's just for them to know for encoding, whether it's stable, so it's needed in the encoding, also I updated InfusionData to just be UniswapV2 data, as they don't use new param

aburkut (approved):

Wertusser (approved):

PR #761, opened 2024-08-07
BACK-1579: V5 vs V6 support separation

JackieJoo:


validate.ts
KanievskyiDanylo, line 1: Probably we don't this file in the code-base, as it's single time update

0xNazarii (approved):

KanievskyiDanylo (approved):

PR #782, opened 2024-09-04
fix(BACK-1691): consider base token oracle state feasibility for WooFiV2

0xNazarii:


src/dex/woo-fi-v2/woo-fi-v2-math.ts
```diff
@@ -97,6 +113,14 @@ export class WooFiV2Math {
       let state1: TokenState = this.state.tokenStates[baseToken1];
       let state2: TokenState = this.state.tokenStates[baseToken2];
 
+      if (!state1.woFeasible || !state2.woFeasible) {
+        const token = !state1.woFeasible ? baseToken1 : baseToken2;
+        this.logger.error(
+          `Oracle state for base token ${token} is not feasible`,
+        );
+        return null;
```
KanievskyiDanylo, line 121: Both tokens can have `woFeasible` false, would be nice to log all of them

aburkut (approved):

src/dex/woo-fi-v2/woo-fi-v2-math.ts
```diff
@@ -46,6 +46,14 @@ export class WooFiV2Math {
   _querySellBase(baseToken: Address, baseAmounts: bigint[]): bigint[] | null {
     try {
       const _state = this.state.tokenStates[baseToken];
+
+      if (!_state.woFeasible) {
+        this.logger.error(
```
KanievskyiDanylo, line 51: Can we also log `this.state.oracleTimestamp` to ease debugging for such cases?

KanievskyiDanylo (approved):

src/dex/woo-fi-v2/woo-fi-v2-math.ts
0xNazarii: done

src/dex/woo-fi-v2/woo-fi-v2-math.ts
0xNazarii: done

PR #799, opened 2024-09-30
BACK-1701: Integrate Bebop

JackieJoo:


src/dex/bebop/bebop.ts
```diff
@@ -0,0 +1,744 @@
...
+          },
+        },
+      },
+    );
+  }
+
+  async initializePricing(blockNumber: number) {
+    if (!this.dexHelper.config.isSlave) {
+      this.rateFetcher.start();
+      await sleep(BEBOP_INIT_TIMEOUT_MS);
```
aburkut, line 125: @JackieJoo Could you please explain why we need this timeout here?

src/dex/bebop/bebop.ts
```diff
@@ -0,0 +1,744 @@
...
+    };
+  }
+
+  async preProcessTransaction(
+    optimalSwapExchange: OptimalSwapExchange<BebopData>,
+    srcToken: Token,
+    destToken: Token,
+    side: SwapSide,
+    options: PreprocessTransactionOptions,
+  ): Promise<[OptimalSwapExchange<BebopData>, ExchangeTxInfo]> {
```
aburkut, line 606: @JackieJoo I don't see pools restriction logic in preProcessTransaction. Usually we have this logic for RFQs

src/dex/bebop/bebop.ts
```diff
@@ -0,0 +1,744 @@
...
+    side: SwapSide,
+    blockNumber: number,
+    limitPools?: string[],
+  ): Promise<null | ExchangePrices<BebopData>> {
+    this.tokensMap = (await this.getCachedTokens()) || {};
+
+    try {
+      const pools =
+        limitPools ??
+        (await this.getPoolIdentifiers(srcToken, destToken, side, blockNumber));
```
KanievskyiDanylo, line 365: Here safer to filter over limitPools if available, e.g. `pools = limitPools.filter(p => <pool is available>)`

src/dex/bebop/bebop.ts
```diff
@@ -0,0 +1,744 @@
...
+      const outDecimals = SwapSide.SELL
+        ? destToken.decimals
+        : srcToken.decimals;
+
+      return [
+        {
+          prices: outputs,
+          unit: BigInt(outDecimals),
+          data: {},
+          poolIdentifier: pools[0],
```
KanievskyiDanylo, line 402: maybe better `instructions.pair` ? cause looks like pools are not used in the pricing

src/dex/bebop/validators.ts
```diff
@@ -0,0 +1,43 @@
+import joi from 'joi';
+
+// const pairValidator = joi.object({
```
KanievskyiDanylo, line 3: Let's remove redundant stuff

src/lib/fetcher/wsFetcher.ts
```diff
@@ -0,0 +1,138 @@
...
+      this.startPolling();
+    }, 3000);
+  }
+
+  private onError(error: any) {
+    this.logger.error(`Connection Error: ${error.toString()}. Reconnecting...`);
+    // reconnect on errors / failures
+    setTimeout(() => {
+      this.startPolling();
+    }, 3000);
```
KanievskyiDanylo, line 69: Does prev connection close before we start new one? it could leak to too many connections

src/lib/fetcher/wsFetcher.ts
```diff
@@ -0,0 +1,138 @@
...
+    const authorization =
+      this.requests.info.requestOptions.headers!.authorization;
+    const name = this.requests.info.requestOptions.headers!.name;
+    if (typeof authorization !== 'string') {
+      throw new Error('Authorization header is not a string');
+    }
+    if (typeof name !== 'string') {
+      throw new Error('Name header is not a string');
+    }
+    this.ws.connect(
```
KanievskyiDanylo, line 106: Better to add log here about attempt to connect

src/dex/bebop/bebop.ts
```diff
@@ -0,0 +1,744 @@
...
+  // update common state required for calculating
+  // getTopPoolsForToken. It is optional for a DEX
+  // to implement this
+  async updatePoolState(): Promise<void> {
+    const tokens = await this.getCachedTokens();
+
+    if (tokens) {
+      this.tokensMap = tokens;
+    }
+  }
```
KanievskyiDanylo, line 499: Not sure that we need this, if we always get data from cache on `getTopPoolsForToken`

src/dex/bebop/bebop.ts
```diff
@@ -0,0 +1,744 @@
...
+        liquidityUSD = liquidityInBase * baseTokenUsd;
+      }
+      if (liquidityUSD) {
+        if (pools.length === 0) {
+          pools.push({
+            exchange: this.dexKey,
+            address: this.settlementAddress,
+            connectorTokens: [
+              {
+                address: quote,
```
KanievskyiDanylo, line 555: Shouldn't it be `quote` or `base` token based on the direction?

src/dex/bebop/bebop.ts
```diff
@@ -0,0 +1,744 @@
...
+    const isSell = side === SwapSide.SELL;
+    const isBuy = side === SwapSide.BUY;
+
+    const params = {
+      sell_tokens: utils.getAddress(srcToken.address),
+      buy_tokens: utils.getAddress(destToken.address),
+      sell_amounts: isSell ? optimalSwapExchange.srcAmount : undefined,
+      buy_amounts: isBuy ? optimalSwapExchange.destAmount : undefined,
+      taker_address: utils.getAddress(options.executionContractAddress),
+      receiver_address: utils.getAddress(options.executionContractAddress),
```
KanievskyiDanylo, line 616: shoudn't it be `options.recipient`?

src/dex/bebop/bebop.ts
JackieJoo: it a websocket connection, so we don't quite know when there's initial state polling finished. I don't think we need to add some complex logic for event emitting on initial state. A buffer of around 3 secs is enough. Just to make sure that the data is in cache. Overall it's mostly for testing, because in reality, we'll just see pricing after a few seconds after initial deployment.

src/dex/bebop/bebop.ts
JackieJoo: I'd rather leave what was calculated by `getPoolIdentifiers`

src/dex/bebop/bebop.ts
```diff
@@ -0,0 +1,744 @@
...
+  BEBOP_WS_API_URL,
+} from './constants';
+import BigNumber from 'bignumber.js';
+import { getBigNumberPow } from '../../bignumber-constants';
+import { utils } from 'ethers';
+import qs from 'qs';
+
+export class Bebop extends SimpleExchange implements IDex<BebopData> {
+  readonly hasConstantPriceLargeAmounts = false;
+  readonly needWrapNative = false;
```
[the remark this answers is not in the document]
JackieJoo: yes, they support eth

src/config.ts
```diff
@@ -151,6 +153,7 @@ const baseConfigs: { [network: number]: BaseConfig } = {
       Executor03: '0x20004f017a0bC0050bc004d9C500a7A089800000',
     },
     hashFlowAuthToken: process.env.API_KEY_HASHFLOW_AUTH_TOKEN || '',
+    bebopAuthToken: process.env.API_KEY_BEBOP_AUTH_TOKEN || '',
```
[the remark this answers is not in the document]
JackieJoo: Currently no

src/lib/fetcher/wsFetcher.ts
JackieJoo: added

src/lib/fetcher/wsFetcher.ts
JackieJoo: added close before trying to reconnect

src/dex/bebop/validators.ts
JackieJoo: done

src/dex/bebop/bebop.ts
JackieJoo: done

src/dex/bebop/bebop.ts
```diff
@@ -0,0 +1,744 @@
...
+  // any pools can be used.
+  async getPricesVolume(
+    srcToken: Token,
+    destToken: Token,
+    amounts: bigint[],
+    side: SwapSide,
+    blockNumber: number,
+    limitPools?: string[],
+  ): Promise<null | ExchangePrices<BebopData>> {
+    this.tokensMap = (await this.getCachedTokens()) || {};
```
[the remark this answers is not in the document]
JackieJoo: removed

src/dex/bebop/bebop.ts
JackieJoo: updated

src/dex/bebop/bebop.ts
JackieJoo: updated

src/dex/bebop/bebop.ts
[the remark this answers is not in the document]
KanievskyiDanylo: Question was related to wrapping/unwrapping, but I checked as well. They do so on their contract, so no additional steps required on our side

src/config.ts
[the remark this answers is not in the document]
KanievskyiDanylo: I prefer not to put redandant values then, we can add this when it's needed

src/dex/bebop/bebop.ts
KanievskyiDanylo: We set tokens for each `tokensIntervalMs` in redis from init service. Don't really understand why we need to read tokens twice if we can read only in `getTopPoolsForToken`

KanievskyiDanylo: I think it's easier to get tokens from cache `getTopPoolsForToken` instead of using `updatePoolState` (might have less redis calls in general then), but current implementation should also work

src/config.ts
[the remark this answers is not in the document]
JackieJoo: done

src/config.ts
```diff
@@ -221,6 +223,7 @@ const baseConfigs: { [network: number]: BaseConfig } = {
     hashFlowDisabledMMs:
       process.env[`HASHFLOW_DISABLED_MMS_43114`]?.split(',') || [],
     dexalotAuthToken: process.env.API_KEY_DEXALOT_AUTH_TOKEN || '',
+    bebopAuthToken: process.env.API_KEY_BEBOP_AUTH_TOKEN || '',
```
KanievskyiDanylo, line 226: Bebop integration is not available on Avalanche at the moment

aburkut: @KanievskyiDanylo removed

aburkut (approved):

PR #804, opened 2024-10-07
GenericRFQ tests fixes

KanievskyiDanylo:
- **chore: add tests setup for generic-rfq e2e**
- **fix: handle false `SKIP_TENDERLY` env & add smart tokens for WETH&USDC on Arbitrum**
- **chore: clean e2e-test-config file**
- **fix: use AugustusV6 address for newTestE2E**
- **fix: use correct balance state field for USDC on Arbitrum**

Fixes BACK-1674


src/dex/generic-rfq/generic-rfq-e2e.test.ts
```diff
@@ -85,7 +85,7 @@ const buildConfigForGenericRFQ = (): RFQConfig => {
   };
 };
 
-const SKIP_TENDERLY = !!getEnv('GENERIC_RFQ_SKIP_TENDERLY', true);
+const SKIP_TENDERLY = getEnv('GENERIC_RFQ_SKIP_TENDERLY', true) === 'true';
```
0xNazarii, line 88: this change will make `SKIP_TENDERLY` `false` in case when `GENERIC_RFQ_SKIP_TENDERLY` is not present in `.env`
we could make the default `true` a string to accommodate

0xNazarii (approved):

src/dex/generic-rfq/generic-rfq-e2e.test.ts
KanievskyiDanylo: updated

PR #807, opened 2024-10-07
Update Hashflow to be needWrapNative=true + CurveV1Factory dynamic needWrapNative

aburkut:
### Summary of the changes
- [x] Make Hashflow needWrapNative=true
- [x] Introduce dynamic needWrapNative into the system, needWrapNative could be a function which dynamically returns boolean
- [x] Make CurveV1Factory to support dynamic needWrapNative
- [x] partialSwap fix for Dexalot


src/implementations/local-paraswap-sdk.ts
```diff
@@ -200,11 +200,12 @@ export class LocalParaswapSDK implements IParaSwapSDK {
       others: [],
       side,
       // For V5 tests, put Augustus V5 address here
-      // contractAddress: '0xDEF171Fe48CF0115B1d80b88dc8eAB59176FEe57',
-      contractAddress: '',
+      contractAddress: '0xDEF171Fe48CF0115B1d80b88dc8eAB59176FEe57',
```
JackieJoo, line 203: Can you leave the implementation to be working on V6?

src/dex/aave-v2/aave-v2.ts
```diff
@@ -51,6 +51,8 @@ export class AaveV2
   readonly hasConstantPriceLargeAmounts = true;
   readonly isFeeOnTransferSupported = false;
 
+  needWrapNative = false;
```
JackieJoo, line 54: fyi you don't need to explicitly specify `false`, as it's `false` by default - https://github.com/paraswap/paraswap-dex-lib/blob/master/src/dex/simple-exchange.ts#L41

src/dex/curve-v1-factory/curve-v1-factory.ts
```diff
@@ -205,6 +223,33 @@ export class CurveV1Factory
...
   }
 
+  private _needWrapNative(
+    srcToken: Address,
+    destToken: Address,
+    data: CurveV1FactoryData,
+  ): boolean {
+    const wethAddress =
+      this.dexHelper.config.data.wrappedNativeTokenAddress.toLowerCase();
+    const ethAddress = ETHER_ADDRESS.toLowerCase();
```
KanievskyiDanylo, line 233: `ETHER_ADDRESS` is already in lower

src/implementations/local-paraswap-sdk.ts
aburkut: done

src/dex/aave-v2/aave-v2.ts
aburkut: refactored needWrapNative a little bit, now it's possible to use default value false

src/dex/curve-v1-factory/curve-v1-factory.ts
aburkut: fixed

JackieJoo (approved):

PR #824, opened 2024-11-01
update FluidDex to use single instance of stateful event subscriber

aburkut:


src/dex/fluid-dex/fluid-dex-liquidity-proxy.ts
```diff
@@ -51,30 +54,21 @@ export class FluidDexEventPool extends StatefulEventSubscriber<FluidDexPoolState
...
-  ): Promise<DeepReadonly<FluidDexPoolState> | null> {
-    if (!(event.args.user in [this.pool])) {
-      return null;
-    }
+  ): Promise<DeepReadonly<FluidDexLiquidityProxyState> | null> {
     const resolverContract = new Contract(
       this.commonAddresses.resolver,
       ResolverABI,
       this.dexHelper.provider,
     );
```
KanievskyiDanylo, line 64: We can create this contract only once in constructor

src/dex/fluid-dex/fluid-dex-liquidity-proxy.ts
```diff
@@ -51,30 +54,21 @@ export class FluidDexEventPool extends StatefulEventSubscriber<FluidDexPoolState
...
+        blockTag: log.blockNumber,
+      });
 
-    this.setState(
-      generatedState,
-      await this.dexHelper.provider.getBlockNumber(),
-    );
+    const generatedState = this.convertToFluidDexPoolState(rawResult);
 
     return generatedState;
```
KanievskyiDanylo, line 73: Can we replace it to `return this.generateState(log.blockNumber)`?

aburkut: @KanievskyiDanylo updated

src/dex/fluid-dex/fluid-dex.ts
```diff
@@ -60,12 +61,21 @@ export class FluidDex extends SimpleExchange implements IDex<FluidDexData> {
     super(dexHelper, dexKey);
     this.logger = dexHelper.getLogger(dexKey);
     this.factory = new FluidDexFactory(
```
KanievskyiDanylo, line 63: Am I right, that FluidDexFactory doesn't notify in any way that new pool was deployed? So, techically we won't see pool until `liqudityProxy` invokes `LogOperate` event?

src/dex/fluid-dex/fluid-dex-liquidity-proxy.ts
aburkut: @KanievskyiDanylo solved

src/dex/fluid-dex/fluid-dex.ts
```diff
@@ -198,26 +191,28 @@ export class FluidDex extends SimpleExchange implements IDex<FluidDexData> {
       // Make sure the pool meets the optional limitPools filter.
       if (limitPools && !limitPools.includes(pool.id)) return null;
 
-      const eventPool = this.eventPools[pool.id];
-
-      if (!eventPool) {
-        this.logger.error(`fluid-dex pool ${pool.id}: No EventPool found.`);
+      const liquidityProxyState = this.liquidityProxy.getState(blockNumber);
```
KanievskyiDanylo, line 194: Maybe better `getStateOrGenerate()`? Don't really see in what cases it can worse anything (it can overuse RPC calls instead of return empty prices when state is null, but to detect this better adding log that we generating state)

src/dex/fluid-dex/fluid-dex.ts
```diff
@@ -19,23 +19,22 @@ import {
...
 import { FluidDexFactory } from './fluid-dex-factory';
 import { getDexKeysWithNetwork, getBigIntPow } from '../../utils';
 import { extractReturnAmountPosition } from '../../executor/utils';
 import { MultiResult } from '../../lib/multi-wrapper';
 import { generalDecoder } from '../../lib/decoders';
 import { BigNumber } from 'ethers';
 import { sqrt } from './utils';
+import { FluidDexLiquidityProxy } from './fluid-dex-liquidity-proxy';
 
 export class FluidDex extends SimpleExchange implements IDex<FluidDexData> {
```
KanievskyiDanylo, line 37: btw, should we support `deposit/withdraw` of fUSDC token in this dex integration? (like in [this transaction)](0x844c04c38ad3b8853238b735ee0eaf6b96df04dbe8911153b0d81eeb8e9a5125))

src/dex/fluid-dex/fluid-dex.ts
KanievskyiDanylo: even then we won't use this pool, becase `this.pools` field is not updated. So, maybe we should update `this.pools` in factory during events handling

src/dex/fluid-dex/fluid-dex.ts
aburkut: I think no. I can see that LogOperate event was triggered within this transaction

src/dex/fluid-dex/fluid-dex.ts
aburkut: @KanievskyiDanylo fixed

KanievskyiDanylo (approved):

0xNazarii (approved):

PR #829, opened 2024-11-06
Feat/back 1700 fix

Birua:
Integration tests are updated to use checkPoolPrices function to actually check amounts.

Please review preProcessTransaction logic because
flag `specialDexSupportsInsertFromAmount: false` has no effect and the signature check is failing.


aburkut:
Was merged into https://github.com/paraswap/paraswap-dex-lib/pull/831

PR #831, opened 2024-11-12
Integrate Cables

aburkut:
Original PR - https://github.com/paraswap/paraswap-dex-lib/pull/788


src/dex/cables/cables.ts
```diff
@@ -0,0 +1,888 @@
...
+        return acc;
+      }, {});
+    }
+  }
+
+  async getTopPoolsForToken(
+    tokenAddress: Address,
+    limit: number,
+  ): Promise<PoolLiquidity[]> {
+    return [];
```
JackieJoo, line 693: Please implement

src/dex/dexalot/dexalot.ts
```diff
@@ -1035,56 +1036,7 @@ export class Dexalot extends SimpleExchange implements IDex<DexalotData> {
     tokenAddress: Address,
     limit: number,
   ): Promise<PoolLiquidity[]> {
-    const normalizedTokenAddress = this.normalizeAddress(tokenAddress);
```
JackieJoo, line 1038: Please return back

JackieJoo (approved):

src/dex/cables/cables.ts
aburkut: implemented

src/dex/cables/cables.ts
```diff
@@ -0,0 +1,1006 @@
...
+      }
+    }
+    return null;
+  }
+
+  async isBlacklisted(txOrigin: Address): Promise<boolean> {
+    const cachedBlacklist = await this.dexHelper.cache.get(
+      this.dexKey,
+      this.network,
+      CABLES_BLACKLIST_CACHE_KEY,
```
KanievskyiDanylo, line 923: Could it be replaced to `this.rateFetcher.blacklistCacheKey` so we don't need separate `CABLES_BLACKLIST_CACHE_KEY` variable?

src/dex/cables/cables.ts
```diff
@@ -0,0 +1,1006 @@
...
+    return address.toLowerCase();
+  }
+
+  getTokenFromAddress(address: Address): Token {
+    return this.tokensMap[this.normalizeAddress(address)];
+  }
+
+  getPoolIdentifier(srcAddress: Address, destAddress: Address, mm?: string) {
+    return `${this.dexKey}_${srcAddress}_${destAddress}`.toLowerCase();
+    return `${this.dexKey}_${srcAddress}_${destAddress}_${mm}`.toLowerCase();
```
KanievskyiDanylo, line 393: looks like first `return` is redundant

src/dex/cables/cables.ts
```diff
@@ -0,0 +1,1006 @@
...
+    }
+    return;
+  }
+
+  getAdapters(side: SwapSide): { name: string; index: number }[] | null {
+    return null;
+  }
+
+  releaseResources?(): AsyncOrSync<void> {
+    if (this.rateFetcher) {
```
KanievskyiDanylo, line 666: looks like `this.rateFetcher` is always present and better to replace `!this.dexHelper.config.isSlave`, just because polling works only on `isSlave = false`

src/dex/cables/cables.ts
```diff
@@ -0,0 +1,1006 @@
...
+          tokensCacheTTLSecs: CABLES_TOKENS_CACHES_TTL_S,
+          tokensCacheKey: 'tokens',
+
+          blacklistIntervalMs: CABLES_API_BLACKLIST_POLLING_INTERVAL_MS,
+          blacklistCacheTTLSecs: CABLES_BLACKLIST_CACHES_TTL_S,
+          blacklistCacheKey: CABLES_BLACKLIST_CACHE_KEY,
+
+          tokensHandleResponseCallback: async () => {
+            await this.setTokensMap();
+          },
```
KanievskyiDanylo, line 122: Why do we need this if it's used only on init service?

src/dex/cables/types.ts
```diff
@@ -0,0 +1,130 @@
...
+}
+
+/**
+ * Utils
+ */
+export type CablesAPIParameters = {
+  url: string;
+  method: Method;
+};
+export class CablesRfqError extends Error {}
```
KanievskyiDanylo, line 27: Unused variables

KanievskyiDanylo, line 27: Unused variables

src/dex/cables/cables.ts
```diff
@@ -0,0 +1,1006 @@
...
+        order: params[0],
+        signature: params[1],
+      },
+    );
+
+    return {
+      targetExchange: this.mainnetRFQAddress,
+      payload,
+      networkFee: '0',
+    };
```
KanievskyiDanylo, line 373: V5 is not supported, we can remove all of that

src/dex/cables/cables.ts
```diff
@@ -0,0 +1,1006 @@
...
+    return `${this.dexKey}_${srcAddress}_${destAddress}_${mm}`.toLowerCase();
+  }
+
+  async getPoolIdentifiers(
+    srcToken: Token,
+    destToken: Token,
+    side: SwapSide,
+    blockNumber: number,
+  ): Promise<string[]> {
+    await this.setTokensMap();
```
KanievskyiDanylo, line 402: let's do Redis cache call after if checks that we do bellow (for empty tokens and same tokens)

src/dex/cables/cables.ts
```diff
@@ -0,0 +1,1006 @@
...
+    }
+
+    const tokensAddr = (await this.getCachedTokensAddr()) || {};
+
+    return [
+      this.getPoolIdentifier(
+        tokensAddr[pairData.base.toLowerCase()],
+        tokensAddr[pairData.quote.toLowerCase()],
+      ),
+    ];
```
KanievskyiDanylo, line 425: In one `getPoolIdentifiers` we do `4` Redis calls. I think it's too much and should be optimized (mostly because there's no sense in making same Redis calls)

src/dex/cables/cables.ts
```diff
@@ -0,0 +1,1006 @@
...
+        }
+      }
+
+      // ---------- Pools ----------
+      let pools = await this.getPoolIdentifiers(
+        srcToken,
+        destToken,
+        side,
+        blockNumber,
+      );
```
KanievskyiDanylo, line 569: We can use passed `limitPools`, no need in making `getPoolIdentifiers` call again

src/dex/cables/cables.ts
```diff
@@ -0,0 +1,1006 @@
...
+      const normalizedSrcToken = this.normalizeToken(srcToken);
+      const normalizedDestToken = this.normalizeToken(destToken);
+      // If: same token, return null
+      if (
+        normalizedSrcToken.address.toLowerCase() ===
+        normalizedDestToken.address.toLowerCase()
+      ) {
+        return null;
+      }
+
```
KanievskyiDanylo, line 550: Let's do this early return before any Redis calls

src/dex/cables/cables.ts
```diff
@@ -0,0 +1,1006 @@
...
+      for (const symbol of Object.keys(this.tokensMap)) {
```
KanievskyiDanylo, line 551: According to the interface 
```ts
  private tokensMap: { [address: string]: Token } = {};
```

we use `address` as key in this `tokensMap`, therefore, does it make sense to iterate over this map, if we can just get tokens by addresses?

src/dex/cables/cables.ts
```diff
@@ -0,0 +1,1006 @@
...
+    if (tokens) {
+      this.tokensMap = Object.keys(tokens).reduce((acc, key) => {
+        //@ts-ignore
+        acc[tokens[key].address.toLowerCase()] = tokens[key];
+        return acc;
+      }, {});
+    }
+  }
+
+  async getTopPoolsForToken(
```
KanievskyiDanylo, line 689: If this integration is not going to introduce new token pairs as we already have on other dexs, I prefer to just return empty array. Otherwise, we are going to too much multiple Redis, RPC, usdTokenPrice calls. So, imho, it should be either emptified or simplified.

src/dex/cables/cables.ts
aburkut: the second is redundant

src/dex/cables/cables.ts
aburkut: fixed

src/dex/cables/cables.ts
aburkut: removed tokensHandleResponseCallback

src/dex/cables/types.ts
```diff
@@ -0,0 +1,130 @@
...
+  [network: string]: {
+    [token: string]: Token;
+  };
+};
+
+export type PairsDataMap = {
+  [network: string]: {
+    [token: string]: PairData;
+  };
+};
```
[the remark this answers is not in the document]
aburkut: fixed

src/dex/cables/types.ts
aburkut: fixed

src/dex/cables/cables.ts
aburkut: removed

src/dex/cables/cables.ts
aburkut: fixed

src/dex/cables/cables.ts
aburkut: fixed

src/dex/cables/cables.ts
aburkut: fixed

src/dex/cables/cables.ts
aburkut: fixed

src/dex/cables/cables.ts
aburkut: fixed

src/dex/cables/cables.ts
aburkut: fixed

KanievskyiDanylo (approved):

PR #832, opened 2024-11-12
remove useless data for FluidDex

aburkut:


KanievskyiDanylo (approved):

0xNazarii (approved):

src/dex/fluid-dex/fluid-dex.ts
```diff
@@ -154,10 +154,10 @@ export class FluidDex extends SimpleExchange implements IDex<FluidDexData> {
     return pool ? [pool.id] : [];
```
KanievskyiDanylo, line 154: in `getPoolIdentifiers` we can get rid of `await` before `getPoolByTokenPair` call as well

aburkut: @KanievskyiDanylo fixed

PR #836, opened 2024-11-19
feat: balancer v2 buy support for specific pools

0xNazarii:


aburkut:
Closed as not actual anymore since buy support was implemented within PR https://github.com/paraswap/paraswap-dex-lib/pull/837

PR #845, opened 2024-12-03
BACK-1776: Update to support node v20 & typescript v5

JackieJoo:


aburkut (approved):

src/dex/algebra/lib/AlgebraMath.ts
```diff
@@ -155,7 +155,7 @@ function _priceComputationCycles(
         ? zeroForOne
           ? poolState.globalState.feeZto
           : poolState.globalState.feeOtz
-        : poolState.globalState.fee,
+        : (poolState.globalState as any).fee,
```
0xNazarii, line 158: we can get rid of `any` by wrapping `isPoolV1_9` typeguard params in `DeepReadonly` generic and changing `poolState` param of `_priceComputationCycles` function from `DeepReadonly<PoolStateV1_1 | PoolState_v1_9>` to `DeepReadonly<PoolStateV1_1> | DeepReadonly<PoolState_v1_9>`

0xNazarii: Its not a logic change, just a typing update and a very simple one
and lets be real if we leave it as that its never going to be fixed

JackieJoo: updated

0xNazarii (approved):

PR #849, opened 2024-12-06
Aburkut/back 1780 stader

aburkut:
Original PR https://github.com/paraswap/paraswap-dex-lib/pull/848


JackieJoo (approved):

src/dex/stader/stader.ts
```diff
@@ -0,0 +1,240 @@
...
+    blockNumber: number,
+    limitPools?: string[] | undefined,
+    transferFees?: TransferFeeParams | undefined,
+    isFirstSwap?: boolean | undefined,
+  ): Promise<ExchangePrices<StaderData> | null> {
+    if (side === SwapSide.BUY) return null;
+    if (!this.isEligibleSwap(srcToken, destToken)) return null;
+
+    const pool = this.ethxPool;
+    if (!pool.getState(blockNumber)) return null;
```
KanievskyiDanylo, line 110: Does it make sense to implement `getOrGenerateState` on this step? I think we won't lose anything, as this dex is used only for specific token pair

aburkut: I think we can

KanievskyiDanylo (approved):

src/dex/stader/stader.ts
aburkut: @KanievskyiDanylo fixed

PR #856, opened 2024-12-11
feat: aave stata v2

sakulstra:
The aave DAO has recently released the next revision of 4626 aToken wrappers: https://x.com/MixBytes/status/1866840081362784603

The new version is not an upgrade but a new deployment and thus needs a new integration.
While the implementations are very similar on the surface there are some small details that changed:
- `factory. getStaticATokens()` is now `factory.getStataTokens()`
- `deposit` and `redeem` are no longer overloaded, but instead a new `redeemATokens` and `depositATokens` was added

I did not manage to run the tests.


src/dex/aave-gsm/config.ts
```diff
@@ -1,13 +1,13 @@
 import { DexParams } from './types';
 import { DexConfigMap } from '../../types';
 import { Network } from '../../constants';
-import { AaveV3Ethereum, MiscEthereum } from '@bgd-labs/aave-address-book';
+import { AaveV3Ethereum, GhoEthereum } from '@bgd-labs/aave-address-book';
```
sakulstra, line 4: Gho addresses are now in a separated library.

src/dex/aave-v2/tokens.ts
```diff
@@ -1,6 +1,6 @@
 import { aToken, Token } from '../../types';
 import { Network } from '../../constants';
-import { tokenlist } from '@bgd-labs/aave-address-book';
+import tokenlist from '@bgd-labs/aave-address-book/dist/tokenlist';
```
sakulstra, line 3: I didn't manage to make typescript accept this.
On other projects ppl import `@bgd-labs/aave-address-book/tokenlist` just fine, but here i didn't figure out why it does not work. As on v2 for sure no more assets are added could make sense to just copy the list to the repo and drop the import.

src/dex/aave-v3-stata-v2/config.ts
```diff
@@ -0,0 +1,28 @@
...
+  AaveV3Ethereum,
+  AaveV3EthereumLido,
+  AaveV3Gnosis,
+} from '@bgd-labs/aave-address-book';
+
+export const AaveV3StataConfig: DexConfigMap<DexParams> = {
+  AaveV3Stata: {
+    [Network.MAINNET]: {
+      factoryAddresses: [
+        AaveV3Ethereum.STATA_FACTORY,
```
sakulstra, line 14: there is one factory per pool, so can be multiple per network (prev implementation did not support that)

src/dex/aave-v3/config.ts
```diff
@@ -65,12 +64,12 @@ export const Config: DexConfigMap<DexParam> = {
       poolAddress: AaveV3BNB.POOL,
       wethGatewayAddress: AaveV3BNB.WETH_GATEWAY,
     },
-    [Network.ZKEVM]: {
```
sakulstra, line 68: zkEVM was never activated.

src/dex/stkgho/config.ts
```diff
@@ -2,13 +2,13 @@ import { DexParams } from './types';
 import { DexConfigMap } from '../../types';
 import { Network } from '../../constants';
 
-import { MiscEthereum } from '@bgd-labs/aave-address-book';
+import { GhoEthereum, AaveSafetyModule } from '@bgd-labs/aave-address-book';
```
sakulstra, line 5: safetymodule now has it's own lib

mounibec:
Resolves BACK-1791

0xNazarii:
@sakulstra It seems like the new `stata` tokens don't have rate like the v1 tokens do, and there appear to be some small differences in the `preview` math, so the current pricing logic won't work.

sakulstra:
@0xNazarii thanks for having a look.

The `rate` is no longer exposed - didn't see that the contract manually replicated the rate logic.
Refactored to use `reserveNormalizedIncome` (which is the rate). How can i reproduce the "small differences in preview" math? From checking the code i don't see why it would be different on this version - perhaps it was related to not fetching the correct rate?

src/dex/aave-v2/tokens.ts
0xNazarii: I don't mind copying the list to the repo, not worth spending time trying to make it work

src/dex/aave-v3-stata-v2/config.ts
```diff
@@ -0,0 +1,30 @@
...
+import { DexConfigMap, AdapterMappings } from '../../types';
+import { Network, SwapSide } from '../../constants';
+import {
+  AaveV3Ethereum,
+  AaveV3EthereumLido,
+  AaveV3Gnosis,
+} from '@bgd-labs/aave-address-book';
+
+export const AaveV3StataConfig: DexConfigMap<DexParams> = {
+  AaveV3Stata: {
```
0xNazarii, line 11:
```suggestion
  AaveV3StataV2: {
```

PR #874, opened 2025-01-07
Added restriction logic

0xprinc:
- Added a new function `restrictPoolWithIds` which adds restricted ids and also filters out existing pools
- temporarily restricted fluid-eth pool


JackieJoo:
changes were merged [here](https://github.com/paraswap/paraswap-dex-lib/pull/875)

PR #877, opened 2025-01-09
Updated Fluid-Dex

0xprinc:
Use pool centerPrice to verify token reserves ratio instead of using the locally calculated pool price. This fixes some issues that can occur because of the previous imprecision here, mainly it fixes the issue why currently the FLUID / ETH pool got hardcoded paused temporarily.
The centerPrice is now returned from the new DexReservesResolver, which is deployed [here](https://github.com/Instadapp/fluid-contracts-public/blob/main/deployments/deployments.md#dexreservesresolver)


0xprinc:
This PR is a WIP right now. Please don't merge it.

0xprinc:
PR is ready to merge.

PR #884, opened 2025-01-17
Support for Balancer V3 hooks with stableSurge hook

johngrantuk:
Adds support for Balancer V3 pools with hooks.

- The aim would be to add any Balancer approved/vetted hooks on a case by case basis.
- Config file is used to keep track of supported hook type and address
- Each hook must implement its own state and event handling
- Added a new event subscriber that will be used to track hook state
- Relevant hook state is used for pool maths and maths repo handles hook logic/calculation if required
- Add stableSurge hook support. See [docs](https://docs.balancer.fi/integration-guides/aggregators/hooks-reference.html#stable-surge-hook) for more info.


src/dex/balancer-v3/hooks/balancer-hook-event-subscriber.ts
```diff
@@ -0,0 +1,226 @@
...
+    logger: Logger,
+  ) {
+    super(parentName, 'Balancer_Hooks', dexHelper, logger);
+
+    // Load supported hooks from config file
+    this.hooksConfigMap = loadHooksConfig(network);
+
+    // Add each hook ABI here
+    this.interfaces = [
+      new Interface(exitFeeHookAbi),
```
KanievskyiDanylo, line 77: seems it's directional hook, not exitHook

src/dex/balancer-v3/hooks/stableSurgeHook.ts
```diff
@@ -0,0 +1,163 @@
...
+      hookInterface,
+      'getSurgeThresholdPercentage',
+      data[dataIndex],
+      'n/a',
+    );
+    const decodedY = decodeThrowError(
+      hookInterface,
+      'getMaxSurgeFeePercentage',
+      data[dataIndex + 1],
+      'n/a',
```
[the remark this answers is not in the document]
aburkut: fixed

src/dex/balancer-v3/balancer-v3-pool.ts
```diff
@@ -186,19 +202,10 @@ export class BalancerV3EventPool extends StatefulEventSubscriber<PoolStateMap> {
       this.interfaces,
     );
```
KanievskyiDanylo, line 204: Don't we need to filter paused pools here?

src/dex/balancer-v3/balancer-v3-pool.ts
```diff
@@ -411,6 +419,12 @@ export class BalancerV3EventPool extends StatefulEventSubscriber<PoolStateMap> {
     let outputAmountRaw = 0n;
     for (const i of indices) {
       const step = steps[i];
+
+      // If the pool has a hook we fetch latest hook state for use in maths
+      let hookState = undefined;
+      if ('hookAddress' in step.poolState && step.poolState.hookAddress)
+        hookState = hookStateMap[step.poolState.hookAddress];
```
KanievskyiDanylo, line 426: What if `hookStateMap[step.poolState.hookAddress][step.pool]` is empty here? for example when new pool with hooks was registered, but `updateHooksState` wasn't invoked yet, so hookStateMap[hook][pool] should be empty

src/dex/balancer-v3/getPoolsApi.ts
```diff
@@ -69,37 +72,53 @@ function createQuery(
...
+          ),
+          weights: pool.poolTokens.map(t =>
+            t.weight ? parseUnits(t.weight, 18).toBigInt() : 0n,
+          ),
+          poolType: pool.type,
+          hookAddress: pool.hook ? pool.hook.address.toLowerCase() : undefined,
+          hookType: pool.hook
+            ? hooksConfigMap[pool.hook.address.toLowerCase()].type
+            : undefined,
+          supportsUnbalancedLiquidity: true,
```
KanievskyiDanylo, line 110: Is it `true` for all supported pools now?

src/dex/balancer-v3/balancer-v3-pool.ts
```diff
@@ -425,14 +439,24 @@ export class BalancerV3EventPool extends StatefulEventSubscriber<PoolStateMap> {
...
           );
         }
+        // StableSurge hook uses Amp as part of maths
+        if (step.poolState.hookType === StableSurge.type && hookState)
+          hookState = {
+            ...(hookState as StableSurgeHookState)[
+              step.poolState.poolAddress.toLowerCase()
+            ],
+            amp: step.poolState.amp,
+          };
```
KanievskyiDanylo, line 449: can this be relevant for non-stable pools?

src/dex/balancer-v3/getPoolsApi.ts
johngrantuk: Its ok to default to true here as this is only required for balancer maths add/remove which we don't use.

src/dex/balancer-v3/balancer-v3-pool.ts
johngrantuk: No. StableSurge is always connected to StablePool (enforced at [factory](https://github.com/balancer/balancer-v3-monorepo/blob/main/pkg/pool-hooks/contracts/StableSurgePoolFactory.sol#L91))

src/dex/balancer-v3/balancer-v3-pool.ts
johngrantuk: Didn't think of that, thanks! In that case we cannot calculate swap result so I've changed to log and return 0.

src/dex/balancer-v3/hooks/stableSurgeHook.ts
```diff
@@ -0,0 +1,193 @@
...
+ */
+export function thresholdSurgePercentageChangedEvent(
+  event: any,
+  state: DeepReadonly<HookStateMap>,
+  log: Readonly<Log>,
+): DeepReadonly<HookStateMap> | null {
+  // ThresholdSurgePercentageChanged(pool, newSurgeThresholdPercentage)
+  const hookAddress = log.address.toLowerCase();
+  if (!state[hookAddress]) {
+    return null;
```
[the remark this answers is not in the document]
aburkut: @KanievskyiDanylo updated

src/dex/balancer-v3/balancer-v3-pool.ts
KanievskyiDanylo: should we return 0n in case `hookStateMap[step.poolState.hookAddress][step.pool]` is empty?

KanievskyiDanylo: image
?

KanievskyiDanylo (approved):

src/dex/balancer-v3/balancer-v3-pool.ts
johngrantuk: Not all hooks will have this pool specific state so I moved the check to the stableSurge specifically for now.

aburkut (approved):

JackieJoo (approved):

PR #886, opened 2025-01-29
test: gas estimation testing

0xNazarii:


tests/tenderly-simulation-new.ts
```diff
@@ -0,0 +1,534 @@
...
+
+    if (normalizedToken === ETHER_ADDRESS) {
+      throw new Error('Cannot provide storage slots for native token');
+    }
+
+    const jsonPath = path.join(__dirname, 'token-storage-slots.json');
+
+    const chainSlots = JSON.parse(
+      await fs.readFile(jsonPath, { encoding: 'utf-8' }),
+    ) as Record<number, Record<string, TokenStorageSlots>>;
```
KanievskyiDanylo, line 508: We can cache it in local memory (and update this cache each time new token is added to json) to prevent `readFile` each test case

tests/utils-e2e.ts
```diff
@@ -1031,3 +1037,147 @@ export const constructE2ETests = (
...
+    amount,
+    swapSide,
+    contractMethod,
+  );
+  console.log({ priceRoute });
+  // prepare state overrides
+  const tenderlySimulator = TenderlySimulatorNew.getInstance();
+  // any address works
+  const userAddress = ethers.Wallet.createRandom().address;
+  // init `StateOverride` object
```
KanievskyiDanylo, line 1071: I suggest to use the same human readable so we used to this. should help us debugging

tests/utils-e2e.ts
```diff
@@ -1031,3 +1037,147 @@ export const constructE2ETests = (
...
+        [userSrcBalanceOfSlot]: ethers.utils.defaultAbiCoder.encode(
+          ['uint'],
+          [amount * 2n],
+        ),
+        [userSrcAllowanceToAugustusSlot]: ethers.utils.defaultAbiCoder.encode(
+          ['uint'],
+          [amount * 2n],
+        ),
+      },
+    };
```
KanievskyiDanylo, line 1109: I think we can move it to separate method in `TenderlySimulatorNew`

tests/tenderly-simulation-new.ts
```diff
@@ -0,0 +1,534 @@
...
+      data: iface.encodeFunctionData('allowance', [owner, spender]),
+      chainId,
+    };
+  }
+
+  async findTokenBalanceOfSlot(
+    chainId: number,
+    token: string,
+  ): Promise<{ slot: string; isVyper?: boolean }> {
+    const account = ethers.constants.AddressZero;
```
KanievskyiDanylo, line 351: I think we can use any non-zero address to make we skip all contract's side logic related to handling ZERO addresses (hard to believe that any contract can handle it separately for `balanceOf` or `allowance` call, but still why not to prevent this)

tests/utils-e2e.ts
0xNazarii: added recognisable names in https://github.com/paraswap/paraswap-dex-lib/pull/886/commits/ed5740ba2baa5a3f78f5d516afa2e8203b22c493

tests/tenderly-simulation-new.ts
0xNazarii: fixed in https://github.com/paraswap/paraswap-dex-lib/pull/886/commits/ed5740ba2baa5a3f78f5d516afa2e8203b22c493, using recognisable names now

tests/tenderly-simulation-new.ts
0xNazarii: added caching in https://github.com/paraswap/paraswap-dex-lib/pull/886/commits/f7780e0cd3b38ce7015ba42b9c4332642d023df3

tests/utils-e2e.ts
0xNazarii: added separate methods in https://github.com/paraswap/paraswap-dex-lib/pull/886/commits/7f7480e01233320cf4cf7c40fd405c2d355ecdf0

KanievskyiDanylo (approved):

PR #887, opened 2025-01-31
Balancer v3 nested erc4626

johngrantuk:
Adds support for nested ERC4626 swaps.

It is possible to have nested ERC4626 tokens in Balancer V3 pools, for example this pool:
https://balancer.fi/pools/ethereum/v3/0x10a04efba5b880e169920fd4348527c64fb29d4d
With:
csUSDC: 0x7204b7dbf9412567835633b6f00c3edc3a8d6330 -- ERC4626 Pool token
  - USDC: 0xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48 -- underlying
csUSDL: 0xbeefc011e94f43b8b7b455ebab290c7ab4e216f1 -- ERC4626 Pool token
  - wUSDL: 0x7751e2f4b8ae93ef6b79d86419d42fe3295a4559 -- ERC4626 underlying
    - USDL: 0xbdc7c08592ee4aa51d06c27ee23d5087d65adbcd -- underlying
The most useful swap paths in this case would be USDL<>USDC. This can be supported by simply adding an extra buffer step to the batch swap to wrap/unwrap as needed.

- Fetch nested ERC4626 addresses from API and include these for token in/out consideration
- Update ERC4626 onchain rate call to include nested
- Handle extra wrap/unwrap steps for nested ERC4626 
- Add integration tests for USDC<>USDL

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@paraswap/dex-lib",
-  "version": "4.0.23",
+  "version": "4.0.24-bal-v3-nested-erc4626.0",
   "main": "build/index.js",
   "types": "build/index.d.ts",
   "repository": "https://github.com/paraswap/paraswap-dex-lib",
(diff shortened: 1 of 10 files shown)


johngrantuk:
I'm closing this as it turns out this ERC4626 has a rebasing underlying which is not compatible with buffers 100% of the time.

PR #904, opened 2025-02-06
fix(fluid-dex): resolve failing e2e tests

SamarendraGouda:


src/dex/fluid-dex/fluid-dex.ts
```diff
@@ -303,6 +293,7 @@ export class FluidDex extends SimpleExchange implements IDex<FluidDexData> {
         }, ${side}:`,
         e,
       );
+      throw e;
```
KanievskyiDanylo, line 296: Returning null should be enough, no need in throwing the error

src/dex/fluid-dex/fluid-dex.ts
```diff
@@ -892,6 +897,7 @@ export class FluidDex extends SimpleExchange implements IDex<FluidDexData> {
       fee,
       outDecimals,
       currentLimits,
+      centerPrice,
       syncTime,
     );
```
KanievskyiDanylo, line 903: should we update `if (amountIn === 2n ** 256n - 1n)` to `if (amountIn === 0n) {` if we decided to return `0n` instead of `2n ** 256n - 1n` ?

src/dex/fluid-dex/fluid-dex-e2e.test.ts
```diff
@@ -283,9 +240,9 @@ const limitsWide: DexLimits = {
   },
 };
 
-const ErrInsufficientBorrowable = new Error('insufficient borrowable');
-const ErrInsufficientMaxPrice = new Error('insufficient max price');
-const ErrInsufficientReserve = new Error('insufficient reserve');
+const ErrInsufficientBorrowable = new Error('DebtLimitReached');
+const ErrInsufficientMaxPrice = new Error('OracleUpdateHugeSwapDiff');
+const ErrInsufficientReserve = new Error('DebtReservesTooLow');
```
KanievskyiDanylo, line 245: assume we should check that `swapInAdjusted` returns `0n` instead of throwing error in tests

JackieJoo: done, those checks were just never done in catch, cause no errors had been thrown

src/dex/fluid-dex/fluid-dex.ts
JackieJoo: done

src/dex/fluid-dex/fluid-dex.ts
JackieJoo: I think it's just dead code, I'll remove the check, cause if we return 0 from swapOutAdjusted, we'll return 0 from swapOut, cause other params don't matter, we'll zero out anyways

JackieJoo: but let me check first

JackieJoo: yeah, max uint should not be returned anyway, but this check doesn't hurt

KanievskyiDanylo: ok, but this check just doesn't make sense anyway

KanievskyiDanylo (approved):

aburkut (approved):

src/dex/fluid-dex/fluid-dex.ts
JackieJoo: removed

PR #906, opened 2025-02-09
BACK-1877: Ekubo Integration

die-herdplatte:
[Ekubo](https://ekubo.org/) is an extensible concentrated liquidity AMM with a singleton architecture.
Initially, the protocol was deployed on Starknet only (where it is the leading AMM) but since February it is also live (in alpha) on Mainnet.
For more information about the current deployment, refer to ["Integrating Ekubo Protocol for aggregators"](https://gist.github.com/moodysalem/636ed2af189100011df48c21c82348b7).

The pricing logic implemented in this PR stems from [Ekubo's TypeScript SDK](https://github.com/EkuboProtocol/starknet-typescript-sdk).

Other links:
- [Docs](https://docs.ekubo.org/)
- [GitHub](https://github.com/EkuboProtocol)
- [Contract addresses](https://docs.ekubo.org/integration-guides/reference/contract-addresses)
- [Discord](https://discord.ekubo.org)


src/dex/ekubo/ekubo.ts
```diff
@@ -0,0 +1,626 @@
...
+  // any pools can be used.
+  async getPricesVolume(
+    srcToken: Token,
+    destToken: Token,
+    amounts: bigint[],
+    side: SwapSide,
+    blockNumber: number,
+    limitPools?: string[],
+  ): Promise<null | ExchangePrices<EkuboData>> {
+    limitPools ??= await this.fetchPools(
```
aburkut, line 207: @die-herdplatte it's not a good idea to make http calls inside getPricesVolume method since it would affect our pricing performance. The better approach would be to `initialize` stateful event subscribers classes (event pools) in `initializePricing` method (for reference please check FluidDex's `initializePricing` method https://github.com/paraswap/paraswap-dex-lib/blob/master/src/dex/fluid-dex/fluid-dex.ts#L114 ). You could use https://eth-mainnet-api.ekubo.org/pools API endpoint to get all pools which could be then used for event pools initialization. Also you could use setInterval (you need to decide which timing interval to use here, it depends on how often you pool's list is updated) inside `initializePricing` method (for reference please check BalancerV3's `initializePricing` method https://github.com/paraswap/paraswap-dex-lib/blob/master/src/dex/balancer-v3/balancer-v3.ts#L97 and don't forget to use clearInteraval in releaseResources method https://github.com/paraswap/paraswap-dex-lib/blob/master/src/dex/balancer-v3/balancer-v3.ts#L634).

die-herdplatte: Ok, I made that decision because the docs say that lazy initialization is preferred.
Anyway, I changed it now as you've proposed.

die-herdplatte:
Since the launch of Ekubo v2 is coming closer (estimated to be live mid March) we decided to not push through with the v1 integration (this PR up until now) and directly integrate v2.
I will implement the v2 integration and then mark this PR as ready for review again.

aburkut:
> Since the launch of Ekubo v2 is coming closer (estimated to be live mid March) we decided to not push through with the v1 integration (this PR up until now) and directly integrate v2. I will implement the v2 integration and then mark this PR as ready for review again.

@die-herdplatte ok

die-herdplatte:
Note that one integration test will fail.
This is due to the current testnet pools not having enough liquidity to support an exact-out swap of 1 ETH.

src/dex/ekubo/ekubo.ts
```diff
@@ -0,0 +1,657 @@
...
+    destToken: Token,
+    _side: SwapSide,
+    blockNumber: number,
+  ): Promise<string[]> {
+    const [token0, token1] = sortAndConvertTokens(srcToken, destToken);
+    const pair = hexStringTokenPair(token0, token1);
+
+    let poolKeys: PoolKey[];
+    try {
+      poolKeys = (await this.fetchAllPoolKeys()).filter(
```
KanievskyiDanylo, line 198: We cannot make this http request each time, in this implementation it should be cached and refreshed with some interval then

src/dex/ekubo/ekubo.ts
```diff
@@ -0,0 +1,657 @@
...
+    const initializedPoolKeys: PoolKey[] = [];
+
+    for (const poolKey of poolKeys) {
+      (this.pools.has(poolKey.string_id)
+        ? initializedPoolKeys
+        : uninitializedPoolKeys
+      ).push(poolKey);
+    }
+
+    const promises = this.initializePools(uninitializedPoolKeys, blockNumber);
```
KanievskyiDanylo, line 240: if pools should be initialized, it should be done on pricing in `getPricesVolume`. This method is designed to only return pool identifiers

KanievskyiDanylo: as I can see, we can initialize pools in refresher job, no need to it on pricing

P.S. we can track and initialize new pools with polling or when Factory emits when new pool is created (if it's the case for this dex)

src/dex/ekubo/ekubo.ts
```diff
@@ -0,0 +1,657 @@
...
+      } catch (err) {
+        this.logger.error('Quote error:', err);
+        continue;
+      }
+    }
+
+    return exchangePrices;
+  }
+
+  // LEGACY
```
KanievskyiDanylo, line 343: not legacy, but for mainnet we can return `NO_PAYLOAD`

src/dex/ekubo/ekubo.ts
KanievskyiDanylo, line 198: Making HTTP call each time is not efficient, let's use cached pools here
(we can also decrease update interval to 1 min)

src/dex/ekubo/ekubo.ts
KanievskyiDanylo, line 240: no need to initialize pools here, we can do it in `updatePoolMap` only

src/dex/ekubo/ekubo.ts
JackieJoo: moved fetching of pool keys to `updatePoolMap` that also stored keys locally and is called periodically (every 1 minute), in all other places we get pool keys from local cache

src/dex/ekubo/ekubo.ts
```diff
@@ -0,0 +1,655 @@
...
+            token0,
+            token1,
+            new PoolConfig(
+              FULL_RANGE_TICK_SPACING,
+              0n,
+              BigInt(this.config.oracle),
+            ),
+          ),
+        );
+      }
```
KanievskyiDanylo, line 226: as we don't fetch from API anything, not sure that this try-catch is still relevant

JackieJoo: you're right, fixed to check for local cache length

src/dex/ekubo/ekubo.ts
JackieJoo: removed initialization here

KanievskyiDanylo (approved):

PR #907, opened 2025-02-10
fix: Use tokenRates as part of balance change in events.

johngrantuk:
* Use token rates as part of scaling


src/dex/balancer-v3/balancer-v3-events.test.ts
```diff
@@ -234,55 +234,22 @@ describe('BalancerV3 EventPool', function () {
     // TODO once we have a new test deployment add tests for: AggregateSwapFeePercentageChanged, SwapFeePercentageChanged, PoolPausedStateChanged
     const eventsToTest: Record<Address, EventMappings> = {
       [BalancerV3Config.BalancerV3[network].vaultAddress]: {
-        // decoding is broken here
         LiquidityAdded: {
-          blockNumbers: [21690379, 21725514, 21725720, 21774068],
```
johngrantuk, line 239: `getPoolTokenRates` fails on earlier block numbers because rateProvider at this time was broken and has since been upgraded to a fixed version.

PR #888, opened 2025-01-31
feat: add M to WrappedM route path

jonalvarezz:
Hi ParaSwap team,

This pull requests adds the M Token -> WrappedM token path. It's done by WrappedM's `wrap`/ `unwrap` call.

* M contract: https://etherscan.io/token/0x866a2bf4e572cbcf37d5071a7a58503bfb36be1b
* WrappedM contract: https://etherscan.io/token/0x437cc33344a0b27a429f795ff6b469c72698b291

WrappedM was added already within `Usual` routes.

Do let us know if anything else is needed. Thanks

**Tests output:**
<details>
  <summary>Click to expand</summary>

BUY: 0x437cc33344a0B27A429f795ff6B469C72698B291 -> 0x866A2BF4E572CbcF37D5071A7a58503Bfb36be1b (swapExactAmountOut)

Screenshot 2025-02-03 at 13 33 48

SELL: 0x866A2BF4E572CbcF37D5071A7a58503Bfb36be1b -> 0x437cc33344a0B27A429f795ff6B469C72698B291 (swapExactAmountIn)

Screenshot 2025-02-03 at 15 04 53

</details>

PR #911, opened 2025-02-20
Aburkut/back 1879 mwrappedm

aburkut:
Original PR https://github.com/paraswap/paraswap-dex-lib/pull/888


src/dex/usual/wrapped-m-m.ts
```diff
@@ -0,0 +1,76 @@
...
+  ): Promise<DexExchangeParam> {
+    if (this.isFromToken(srcToken) && this.isToToken(destToken)) {
+      const fn =
+        srcToken.toLowerCase() === this.config.toToken.address.toLowerCase()
+          ? 'wrap(address, uint256)'
+          : 'unwrap(address, uint256)';
+
+      const exchangeData = this.wrappedMInterface.encodeFunctionData(fn, [
+        recipient,
+        srcAmount,
```
KanievskyiDanylo, line 62: looks like for BUY we should use `destAmount` here (as for swaps are 1:1 and we increase srcAmount with slippage)

KanievskyiDanylo: Please do so on other usual dexs as well

aburkut: ok

KanievskyiDanylo (approved):

JackieJoo (approved):

PR #916, opened 2025-03-05
Balancer api query update

johngrantuk:
Update to use the new Balancer API endpoint: `aggregatorPools`. Its designed to replace `poolGetAggregatorPools`. It should be mostly backward compatible to make it easy to switch but has a new field: `canUseBufferForSwaps`. This will be dynamic, defaulting to false, and will only be changed to true if an ERC4626 token used in a pool has been manually confirmed to:
* be a valid ERC4626 (we check it passes some erc4626 fork tests)
* has an initialised Balancer V3 Buffer
So if canUseBufferForSwaps === true it should always be ok to use to swap.

### DIFF
--- a/package.json
+++ b/package.json
@@ -52,7 +52,7 @@
   },
   "dependencies": {
     "@0x/utils": "^4.5.2",
-    "@balancer-labs/balancer-maths": "^0.0.19",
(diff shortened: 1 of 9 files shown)


KanievskyiDanylo (approved):

KanievskyiDanylo:
Implemented in #919

PR #921, opened 2025-03-10
feat: stata optimism synthetix

sakulstra:


src/dex/aave-v3-stata-v2/config.ts
```diff
@@ -12,29 +12,36 @@ import {
 
 export const AaveV3StataConfig: DexConfigMap<DexParams> = {
   AaveV3StataV2: {
-    [Network.MAINNET]: {
-      factoryAddresses: [
-        AaveV3Ethereum.STATA_FACTORY,
-        AaveV3EthereumLido.STATA_FACTORY,
```
sakulstra, line 18: this was wrong. the factory is per pool and lido is a different pool.

Not sure why on the ui it seems to work: https://app.paraswap.xyz/#/swap/0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2-0xfA1fDbBD71B0aA16162D76914d69cD8CB3Ef92da/999999/SELL?network=ethereum&version=6.2
Perhaps it's not actually working and it's just showing the wrong output token symbol?
As i understand currently deposits would always take the default pool

PR #928, opened 2025-03-14
Add support for Hashflow on Base

gxmxni-hashflow:
- add support for Hashflow on Base
- correct the logic on deriving the price from levels (levels are no longer cumulative)

## Testing
- added unit tests for Base


KanievskyiDanylo:
merged within #931

PR #934, opened 2025-03-19
fix: generate state if null on Spark

KanievskyiDanylo:


src/dex/spark/spark.ts
```diff
@@ -111,7 +111,7 @@ export class Spark
     limitPools?: string[],
   ): Promise<null | ExchangePrices<SparkData>> {
     if (!this.isAppropriatePair(srcToken, destToken)) return null;
-    const state = this.eventPool.getState(blockNumber);
+    const state = this.eventPool.getOrGenerateState(blockNumber);
```
aburkut, line 114: missing await?

PR #951, opened 2025-04-09
fix: Use aggregate swap fee amount in swap/add/remove event updates.

johngrantuk:
Was not taking aggregate swap fee amount into account when updating pool balances. This is now fixed.

SwapFeePercentageChanged event is emitting a slightly different value from the actual setter. This is because the contract is truncating the value before storing and will have a min step of 0.00001% (effectively 5 decimal places of precision).


src/dex/balancer-v3/balancer-v3-e2e.test.ts
```diff
@@ -74,44 +74,44 @@ function testForNetwork(
 
   const sideToContractMethods = new Map([
     [SwapSide.SELL, [ContractMethod.swapExactAmountIn]],
-    [SwapSide.BUY, [ContractMethod.swapExactAmountOut]],
+    // [SwapSide.BUY, [ContractMethod.swapExactAmountOut]],
   ]);
```
Ivan011001, line 78: Do we actually need to comment these test cases?

src/dex/balancer-v3/balancer-v3-e2e.test.ts
```diff
@@ -74,44 +74,44 @@ function testForNetwork(
...
+            //       side === SwapSide.SELL ? tokenAAmount : nativeTokenAmount,
+            //       side,
+            //       dexKey,
+            //       contractMethod,
+            //       network,
+            //       provider,
+            //       poolIdentifiers,
+            //     );
+            //   });
+            // }
```
Ivan011001, line 114: Same question about the need to comment these test cases

src/dex/balancer-v3/balancer-v3-e2e.test.ts
```diff
@@ -126,20 +126,20 @@ function testForNetwork(
...
+            //     holders[tokenBSymbol],
+            //     side === SwapSide.SELL ? tokenBAmount : tokenAAmount,
+            //     side,
+            //     dexKey,
+            //     contractMethod,
+            //     network,
+            //     provider,
+            //     poolIdentifiers,
+            //   );
+            // });
```
Ivan011001, line 142: Same question about the need to comment these test cases

src/dex/balancer-v3/balancer-v3-events.test.ts
```diff
@@ -407,6 +407,136 @@ describe('BalancerV3 EventPool', function () {
     });
   });
 
+  // This test uses a pool with no rates which means it should match 100%
+  describe.only('Mainnet - Test Aggregate Fee (Pool with no rates)', function () {
```
Ivan011001, line 411: Newly added tests are the only that work for me, while previous don't work at all, is this an expected behaviour?

johngrantuk: The previous tests have always failed. This is because the rate can't be tracked by the event system so it won't be updated. There is a timer run to update those.

src/dex/balancer-v3/balancer-v3-e2e.test.ts
johngrantuk: I saw that you guys had commented them earlier and just left as.

src/dex/balancer-v3/balancer-v3-events.test.ts
Ivan011001: Got it, thanks for the explanation

src/dex/balancer-v3/balancer-v3-e2e.test.ts
Ivan011001: OK, I see. @JackieJoo need you opinion regarding this one

src/dex/balancer-v3/balancer-v3-e2e.test.ts
JackieJoo: it's just to only run failed case, uncommented back

src/dex/balancer-v3/balancer-v3-e2e.test.ts
JackieJoo: it's just to only run failed case, uncommented back

Ivan011001 (approved):

PR #954, opened 2025-04-11
fix: Use same precision as SC for SwapFeePercentageChanged event.

johngrantuk:
`SwapFeePercentageChanged` event is emitting a slightly different value from the actual setter. This is because the contract is truncating the value before storing and will have a min step of 0.00001% (effectively 5 decimal places of precision).


johngrantuk:
Combined in #951

PR #960, opened 2025-04-17
Ekubo TWAMM integration

die-herdplatte:
Integrates [Ekubo TWAMM pools](https://docs.ekubo.org/user-guides/dollar-cost-average-orders)

Based on #944

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@paraswap/dex-lib",
-  "version": "4.5.8",
+  "version": "4.5.9",
   "main": "build/index.js",
   "types": "build/index.d.ts",
   "repository": "https://github.com/paraswap/paraswap-dex-lib",


die-herdplatte (commented):
I also saw that you commented out the implementation for `getTopPoolsForToken` added in our last PR.
Could you please explain why you made that decision so that we can potentially fix any issues?

Ivan011001:
> I also saw that you commented out the implementation for `getTopPoolsForToken` added in our last PR. Could you please explain why you made that decision so that we can potentially fix any issues?

Hi @die-herdplatte,

The reason the implementation for `getTopPoolsForToken` commented out is that it attempts to [handle pool initialization directly](https://github.com/die-herdplatte/paraswap-dex-lib/blob/ekubo-twamm/src/dex/ekubo/ekubo.ts#L493-L513). However, this method is called by a service that isn’t responsible for initializing pools, which led to errors

To avoid these issues, you should avoid initializing pools within the `getTopPoolsForToken` method. Still, you can generate the state separately and use it instead. A good example of this approach can be seen in the [Nerve integration](https://github.com/VeloraDEX/paraswap-dex-lib/blob/master/src/dex/nerve/nerve.ts#L76-L106), where the pool can either be initialized or its state can be generated separately depending on the needs

Ivan011001:
Hey @die-herdplatte, could you please add E2E tests for new pools , so we could be sure that it works as expected

die-herdplatte:
> Hey @die-herdplatte, could you please add E2E tests for new pools , so we could be sure that it works as expected

I added the new WBTC - cbBTC pair

src/dex/ekubo/ekubo.ts
```diff
@@ -187,28 +186,28 @@ export class Ekubo extends SimpleExchange implements IDex<EkuboData> {
...
+          token0,
+          token1,
+          new PoolConfig(params.tickSpacing, params.fee, 0n),
+        ),
+        new PoolKey(
+          token0,
+          token1,
+          new PoolConfig(0, params.fee, BigInt(this.config.twamm)),
+        ),
+      ]);
```
Ivan011001: Hey @die-herdplatte, 
Are there any real pools that exist but aren’t returned by the API, and could still be generated locally? We couldn’t find any, so it seems like the logic might be redundant

die-herdplatte: In the best case no, this logic exists as a fallback for when the Ekubo API is down and tries to make a good guess on what suitable pool parameters would be

Ivan011001 (approved):

PR #968, opened 2025-04-24
add console.time for UniV4 pricing

aburkut:


src/dex/uniswap-v4/constants.ts
```diff
@@ -20,6 +20,8 @@ export const POOL_CACHE_REFRESH_INTERVAL = 60 * 60 * 24; // 24 hours
 
 export const SWAP_EVENT_MAX_CYCLES = 10_000;
```
KanievskyiDanylo, line 21: Can we log on swap handling when `SWAP_EVENT_MAX_CYCLES` are reached? as I can see, we don't have this limitation on uni-v3, so maybe this log can reveal some issues on swap event handling

KanievskyiDanylo (approved):

src/dex/uniswap-v4/constants.ts
aburkut: added

KanievskyiDanylo (approved):

PR #973, opened 2025-04-29
Change to use latest state of all pools when new pools are detected

johngrantuk:
Proposed fix for issue reported on 28/04/25. See [this report](https://quark-ceres-740.notion.site/28-04-25-Paraswap-Issue-1e3e395e228180568eb0c5024c683e0a) for more detail on investigation into cause.

TLDR:
Instead of only fetching onchain state for new, previously untracked pools fetch state for all existing pools too. Also update `getOnChainState` to use the specific provider block number. This should ensure that setState is not using an outdated pool state.


0xNazarii (approved):

src/dex/balancer-v3/balancer-v3-pool.ts
```diff
@@ -197,19 +198,16 @@ export class BalancerV3EventPool extends StatefulEventSubscriber<PoolStateMap> {
       return null;
```
KanievskyiDanylo, line 198: as we update states for all pools, not only for the one ones, I guess we should not return here

KanievskyiDanylo (approved):

PR #1010, opened 2025-06-03
Balancer v3 fix step optimisation

johngrantuk:
There is an issue specific to the case of a BalancerV3 exactOut swap type with adjacent buffer steps that form circular swaps. Its only an issue when the Vault does not have balance for the intermediate unwrapped token which hasn't happened until [this case](https://dashboard.tenderly.co/shared/simulation/db83c309-9a74-43a0-bc16-f480f6ea05d3/debugger?trace=0.4.0.2).

We can fix this by removing any adjacent, circular buffer steps buffer steps in `getDexParam` before the swap is encoded. This also has the benefit of making the swap path more gas efficient as it removes unnecessary steps. This PR covers that and adds a e2e test for the case that was previously failing.

### DIFF
--- a/tests/v6/utils-e2e-v6.ts
+++ b/tests/v6/utils-e2e-v6.ts
@@ -55,6 +55,13 @@ export async function runE2ETest(
   const amountToFund = BigInt(srcAmount) * 2n;
   // add allowance override to Augustus
   if (srcToken.toLowerCase() !== ETHER_ADDRESS) {
+    await tenderlySimulator.addTokenBalanceOverride(
+      stateOverride,
+      network,
+      srcToken,
+      senderAddress,
+      amountToFund,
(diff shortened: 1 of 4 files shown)


tests/v6/utils-e2e-v6.ts
```diff
@@ -55,6 +56,16 @@ export async function runE2ETest(
   const amountToFund = BigInt(srcAmount) * 2n;
   // add allowance override to Augustus
   if (srcToken.toLowerCase() !== ETHER_ADDRESS) {
+    if (overrideBalance) {
```
KanievskyiDanylo: Let's add balance by default as it's used for all cases, no need in specifying it separately

0xNazarii (approved):

PR #1012, opened 2025-06-04
feat/active pools

KanievskyiDanylo:
- **temp: add time and iteration count log on uni-v3**
- **4.3.3**

### DIFF
--- a/src/dex/uniswap-v3/uniswap-v3.ts
+++ b/src/dex/uniswap-v3/uniswap-v3.ts
@@ -647,6 +647,8 @@ export class UniswapV3
     blockNumber: number,
     limitPools?: string[],
   ): Promise<null | ExchangePrices<UniswapV3Data>> {
+    const reqId = Math.floor(Math.random() * 10000).toString();
+    const getPricesVolumeStart = Date.now();
     try {
       const _srcToken = this.dexHelper.config.wrapETH(srcToken);
       const _destToken = this.dexHelper.config.wrapETH(destToken);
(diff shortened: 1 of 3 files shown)


Ivan011001 (approved):

src/dex/uniswap-v3/uniswap-v3.ts
```diff
@@ -769,19 +771,59 @@ export class UniswapV3
...
+            (acc, [key, value], index) => {
+              if (value > acc.max) {
+                acc.max = value;
+              }
+              if (value < acc.min) {
+                acc.min = value;
+              }
+              acc.avg += value;
+
+              if (index === Object.entries(iterationCount).length - 1) {
```
0xNazarii: you can just take the array from the callback arguments instead of recreating it

src/dex/uniswap-v3/uniswap-v3.ts
```diff
@@ -769,19 +771,59 @@ export class UniswapV3
...
+
+              return acc;
+            },
+            { avg: 0, max: 0, min: Infinity },
+          );
+
+          this.logger.info(
+            `_getOutputs_${poolKey}_${reqId}: ${
+              getOutputsFinish - getOutputsStart
+            } ms (src: ${srcToken.address}, dest: ${
```
0xNazarii: won't logging tokens be redundant since their addresses are included in the pool key?

KanievskyiDanylo: token0 is not always srcToken, and same for destToken

but I agree that it's a bit redundant, so will just leave src/destTokens in poolKey

0xNazarii (approved):

PR #1014, opened 2025-06-04
feat(BACK-1977): integrate RingV2

Ivan011001:


src/dex/uniswap-v2/ring-v2.ts
```diff
@@ -0,0 +1,368 @@
...
+        isSell ? transferFees.destDexFee : transferFees.srcFee,
+        isSell ? this.DEST_TOKEN_DEX_TRANSFERS : SRC_TOKEN_PARASWAP_TRANSFERS,
+      );
+
+      // As ringv2 just has one pool per token pair
+      return [
+        {
+          prices: outputsWithFee,
+          unit: unitOutWithFee,
+          data: {
```
0xNazarii, line 273: this should only include data which is used for tx building (in `getDexParams()`)

src/dex/uniswap-v2/ring-v2.ts
```diff
@@ -0,0 +1,368 @@
...
+
+    return {
+      needWrapNative: this.needWrapNative,
+      specialDexSupportsInsertFromAmount: true,
+      dexFuncHasRecipient,
+      exchangeData,
+      targetExchange,
+      specialDexFlag,
+      transferSrcTokenBeforeSwap,
+      returnAmountPos: undefined,
```
0xNazarii: as far as I can see, router method returns the amounts

src/dex/uniswap-v2/ring-v2.ts
Ivan011001: We need to return these values in order to satisfy `getPricesVolumes` on UniswapV2 that is extended by RingV2

src/dex/uniswap-v2/ring-v2.ts
```diff
@@ -0,0 +1,371 @@
...
+    }
+
+    const exchangeData = this.exchangeRouterInterface.encodeFunctionData(
+      functionName,
+      args,
+    );
+
+    specialDexFlag = SpecialDex.DEFAULT;
+    targetExchange = data.router;
+    dexFuncHasRecipient = true;
```
KanievskyiDanylo: No need in creating these variables separately, we can simply include them in the response

src/dex/uniswap-v2/ring-v2.ts
```diff
@@ -0,0 +1,371 @@
...
+    let path: Address[] = [from.toLowerCase(), to.toLowerCase()];
+
+    if (isETHAddress(srcToken)) {
+      if (side == SwapSide.SELL) {
+        functionName = RingV2Functions.swapExactETHForTokens;
+        args = [destAmount, path, recipient, deadline];
+      } else {
+        functionName = RingV2Functions.swapETHForExactTokens;
+        args = [srcAmount, path, recipient, deadline];
+      }
```
KanievskyiDanylo: not clear to me why we are passing `destAmount` on SELL instead of `srcAmount`. At the same time, we always handle wrapped tokens on uni-v2, so no need in covering ETH case. Therefore, I suggest to remove ETH related part and leave only `swapExactTokensForTokens` and `swapTokensForExactTokens`

KanievskyiDanylo: also, ETH related methods have [almost no usage](https://bloxy.info/address/0x39d1d8fcC5E6EEAf567Bce4e29B94fec956D3519), so not relevant at least in terms of this

src/dex/uniswap-v2/ring-v2.ts
```diff
@@ -0,0 +1,371 @@
...
+      specialDexSupportsInsertFromAmount: true,
+      dexFuncHasRecipient,
+      exchangeData,
+      targetExchange,
+      specialDexFlag,
+      transferSrcTokenBeforeSwap,
+      returnAmountPos: extractReturnAmountPosition(
+        this.exchangeRouterInterface,
+        functionName,
+        'amounts',
```
KanievskyiDanylo, line 345: I guess we should also pass `outputIndex=1` as we are interested in last amount (amount that is sent to user) ([example](https://dashboard.tenderly.co/tx/0x9c2bf9bcee4be9f7020c335f83afc3a7290355d7a1e4b596474653947713ab95/debugger?trace=0.3.2.2.1.3.0.1.1))

src/dex/uniswap-v2/ring-v2.ts
```diff
@@ -0,0 +1,371 @@
...
+  swapExactETHForTokens = 'swapExactETHForTokens',
+  swapTokensForExactTokens = 'swapTokensForExactTokens',
+  swapTokensForExactETH = 'swapTokensForExactETH',
+  swapETHForExactTokens = 'swapETHForExactTokens',
+}
+
+const RingV2Config: DexConfigMap<DexParams> = {
+  RingV2: {
+    [Network.MAINNET]: {
+      subgraphURL: '2f5DMnspUwMx2n3229koTsFrAZRua3YV69pucVzWmQA9',
```
KanievskyiDanylo: Not sure if PoolTracker is going to work properly, as [subgraph](https://thegraph.com/explorer/subgraphs/2f5DMnspUwMx2n3229koTsFrAZRua3YV69pucVzWmQA9?view=Query&chain=arbitrum-one) returns pairs with fw tokens, so we won't receive original tokens anyway (e.g. DAI, USDC, WETH, etc)

So, I suggest removing subgraph support here to exclude PoolTracker on this dex, or we can also extend PoolTracker method to unwrap fw tokens to original ones (probably not worth it)

src/dex/uniswap-v2/ring-v2.ts
```diff
@@ -0,0 +1,371 @@
...
+    let targetExchange: Address;
+    let dexFuncHasRecipient: boolean;
+
+    let ttl = 1200;
+    const deadline = `0x${(
+      Math.floor(new Date().getTime() / 1000) + ttl
+    ).toString(16)}`;
+
+    const [from, to] = this.getTokenAddresses(srcToken, destToken);
+    let path: Address[] = [from.toLowerCase(), to.toLowerCase()];
```
KanievskyiDanylo: I guess we can use `path` from `data`, therefore no need for input `token: string` support in  `getTokenAddresses`

src/abi/ring-v2/ring-v2-factory.json
KanievskyiDanylo, line 1: looks like this ABI is exactly the same as UniV2 Factory ABI, so no need in duplicate

src/abi/ring-v2/ring-v2-pool.json
KanievskyiDanylo, line 1: looks like this ABI is exactly the same as UniV2 Pool ABI, so no need in duplicate

src/dex/uniswap-v2/ring-v2.ts
```diff
@@ -0,0 +1,349 @@
...
+
+const FewWrappedTokenConfig: DexConfigMap<{
+  fewWrapFactory: string;
+  bytecode: string;
+}> = {
+  RingV2: {
+    [Network.MAINNET]: {
+      fewWrapFactory: '0x7D86394139bf1122E82FDF45Bb4e3b038A4464DD',
+      bytecode: ETHMainnetFewWrappedTokenJSON.bytecode,
+    },
```
KanievskyiDanylo, line 60: Can we simply hardcode `initCode` instead of storing `bytecode` and calculating it each time?

0xNazarii (approved):

PR #1019, opened 2025-06-17
feat: BalV1 encoding fixes + Tenderly state proxy override

KanievskyiDanylo:
- **dev: test price route**
- **test: handle overrides for state proxy tokens**
- **fix: correctly insert from amount on bal-v1**


src/dex/sky-converter/types.ts
```diff
@@ -9,6 +9,6 @@ export type DexParams = {
   newTokenAddress: Address;
   newTokenRateMultiplier: bigint;
   oldToNewFunctionName: string;
-  newToOldFunctionName: string;
+  newToOldFunctionName: string | undefined;
```
0xNazarii: isn't `newToOldFunctionName?: string;` cleaner?

Ivan011001: It is, replaced

src/dex/sky-converter/config.ts
```diff
@@ -23,7 +23,7 @@ export const SkyConverterConfig: DexConfigMap<DexParams> = {
       oldTokenAddress: '0x9f8f72aa9304c8b593d555f12ef6589cc3a579a2',
       newTokenAddress: '0x56072C95FAA701256059aa122697B133aDEd9279',
       oldToNewFunctionName: 'mkrToSky',
-      newToOldFunctionName: 'skyToMkr',
+      newToOldFunctionName: undefined, // no longer possible to convert SKY back to MKR
```
0xNazarii: this line can be removed now

PR #1035, opened 2025-07-15
feat: change new PT to 25 Sept

0xAurelou:
## Description

Changed Old PT from 29 May to 31 July and New PT from 31 Jul to 25 Sep.


Ivan011001:
Handled in #1045

PR #1052, opened 2025-07-29
feat(BACK-2009): integrate BunniV2

Ivan011001:


src/dex/bunni-v2/bunni-v2.ts
```diff
@@ -0,0 +1,670 @@
...
+      return null;
+    }
+
+    const pools = this.getAvailablePoolsForPair(
+      poolStates,
+      srcToken.address,
+      destToken.address,
+    );
+
+    const poolIdSet = new Set((limitPools ?? []).map(id => id));
```
KanievskyiDanylo: redundant map

src/dex/bunni-v2/bunni-v2.ts
```diff
@@ -0,0 +1,670 @@
...
+
+  async getPricesVolume(
+    srcToken: Token,
+    destToken: Token,
+    amounts: bigint[],
+    side: SwapSide,
+    blockNumber: number,
+    limitPools?: string[],
+  ): Promise<null | ExchangePrices<BunniV2Data>> {
+    const poolStates = this.eventPools.getState(blockNumber);
```
KanievskyiDanylo: should we go for RPC pricing in case something went wrong with eventPools EventSubscriber?

probably for that we would need to store initial pool list outside from the state

KanievskyiDanylo: if so, I guess we can do smth like `if (poolStates === null)` then `do getOnChainPrices`

src/dex/bunni-v2/bunni-v2.ts
```diff
@@ -0,0 +1,670 @@
...
+
+    // if block is the chain head, use the chain head timestamp
+    if (blockNumber === this.dexHelper.blockManager.getLatestBlockNumber()) {
+      const activeChainHead = this.dexHelper.blockManager.getActiveChainHead();
+      if (activeChainHead)
+        blockTimestamp = bigIntify(activeChainHead.timestamp);
+    }
+
+    // otherwise, get the block timestamp via RPC call
+    if (blockTimestamp === undefined) {
```
KanievskyiDanylo, line 320: not sure if this possible to enter this case, please add a log here, so we can monitor later

src/dex/bunni-v2/bunni-v2.ts
```diff
@@ -0,0 +1,670 @@
...
+    const funcName =
+      side === SwapSide.SELL
+        ? 'quoteExactInputSingle'
+        : 'quoteExactOutputSingle';
+
+    const callData = amounts.map(amount => {
+      return {
+        target: BunniV2Config.BunniV2[this.network].quoter,
+        callData: this.quoterInterface.encodeFunctionData(funcName, [
+          [pool.key, zeroForOne, amount, '0x'],
```
KanievskyiDanylo: Is it safe to pass `0x` hookData even though pool might have hooks?

src/dex/bunni-v2/bunni-v2.ts
```diff
@@ -0,0 +1,670 @@
...
+          ),
+        });
+      }
+    });
+
+    await updatePricePerVaultShares(
+      this.dexHelper,
+      this.erc4626Interface,
+      pricePerVaultShares,
+    );
```
KanievskyiDanylo: Considering that `pricePerVaultShares` is required for calculating token reserves (if not I would simplify everything), I would suggest to fetch all pricePerVaultShares in updatePoolState method (this method is invoked each several minutes which is sufficient to keep data fresh). Please take a look [at this example](https://github.com/VeloraDEX/paraswap-dex-lib/blob/624acdfdbde62d1ad285e3edd0de34aebe430211/src/dex/woo-fi-v2/woo-fi-v2.ts#L367-L368)

KanievskyiDanylo: It would make only 1 RPC call when `updatePoolState` is invoked, instead of making 1 RPC call for each token

src/dex/bunni-v2/bunni-v2.ts
```diff
@@ -0,0 +1,670 @@
...
+          liquidity1 = totalBalance1 * parseFloat(pool.currency1.price);
+        } else if (parseFloat(pool.currency0.price) > 0) {
+          liquidity1 =
+            totalBalance1 *
+            parseFloat(pool.priceCurrency0) *
+            parseFloat(pool.currency0.price);
+        } else {
+          const usdTokenAmounts = await this.dexHelper.getUsdTokenAmounts([
+            [pool.currency1.id, bigIntify(Math.floor(totalBalance1))],
+          ]);
```
KanievskyiDanylo: You can take all USD tokens values at once with `this.dexHelper.getUsdTokenAmounts()` (instead of making call for each token separately). Here's the [example](https://github.com/VeloraDEX/paraswap-dex-lib/blob/624acdfdbde62d1ad285e3edd0de34aebe430211/src/dex/uniswap-v3/uniswap-v3.ts#L1317)

src/dex/bunni-v2/bunni-v2.ts
```diff
@@ -0,0 +1,670 @@
...
+      network,
+      dexHelper,
+      this.logger,
+    );
+  }
+
+  async initializePricing(blockNumber: number) {
+    await this.eventPools.initialize(blockNumber);
+
+    if (!this.updateVaultSharePricesTimer) {
```
KanievskyiDanylo, line 100: let's add this interval only when `this.dexHelper.config.isSlave === true`

src/dex/bunni-v2/bunni-v2.ts
```diff
@@ -0,0 +1,670 @@
...
+    side: SwapSide,
+    blockNumber: number,
+    limitPools?: string[],
+  ): Promise<null | ExchangePrices<BunniV2Data>> {
+    const poolStates = this.eventPools.getState(blockNumber);
+    if (poolStates === null) return null;
+
+    const _from = this.dexHelper.config.wrapETH(srcToken);
+    const _to = this.dexHelper.config.wrapETH(destToken);
+    if (_from.address === _to.address) {
```
KanievskyiDanylo: let's do early return if `limitPools` is empty

src/dex/bunni-v2/bunni-v2.ts
0xSzeth: All pools in our dex have hooks, but none of the require that we pass hookData. Passing `0x` is always sufficient for the Bunni v2 dex.

src/dex/bunni-v2/bunni-v2.ts
0xSzeth: How about a subgraph based method? If `poolStates === null` we can fetch available pools for the token pair from the subgraph, and then use RPC pricing.

src/dex/bunni-v2/bunni-v2.ts
0xSzeth: Fixed in PR 1056 commit f540024

src/dex/bunni-v2/bunni-v2.ts
0xSzeth: Added in PR 1056 commit db81c06

src/dex/bunni-v2/bunni-v2.ts
0xSzeth: Added in PR 1056 commit 6a9f11e

src/dex/bunni-v2/bunni-v2.ts
0xSzeth: We are already updating vault share prices on the event pool every 60 seconds, which is needed for other event handlers. So I've made some improvements there for use in the `getTopPoolsForToken()` function and removed the existing logic as it is redundant.

Please see PR 1056 commit 882154c

src/dex/bunni-v2/bunni-v2.ts
0xSzeth: Fixed in PR 1056 commit 1b40c8c

src/dex/bunni-v2/bunni-v2.ts
0xSzeth: Added in PR 1056 commit e65d080

src/dex/bunni-v2/bunni-v2.ts
0xSzeth: Fixed in PR 1056 commit cb76ff2

KanievskyiDanylo (approved):

PR #1053, opened 2025-07-29
DRAFT: Add Blackhole CL DEX integration with configuration and E2E tests

sandeepuserblitz:


KanievskyiDanylo:
@claude

src/dex/algebra-integral/forks/blackhole-cl.ts
```diff
@@ -0,0 +1,194 @@
...
+      _isSrcTokenTransferFeeToBeExchanged
+        ? applyTransferFee(
+            poolAmounts,
+            side,
+            transferFees.srcDexFee,
+            SRC_TOKEN_DEX_TRANSFERS,
+          )
+        : poolAmounts,
+    );
+
```
KanievskyiDanylo: Can we combine `balances` and `rpcPrices` calls in one single multicall? would be a nice improvement to go from 2RPC to 1RPC

src/dex/algebra-integral/forks/blackhole-cl.ts
```diff
@@ -0,0 +1,194 @@
...
+              ['uint256', 'uint256'], // amountOut, amountIn
+              [0n, 0n],
+              result => result.map((amount: BigNumber) => amount.toBigInt()),
+            );
+
+            return isSELL ? parsed[0] : parsed[1];
+          },
+        }));
+    });
+
```
KanievskyiDanylo: Also, as I can see, the only [difference](https://www.diffchecker.com/gdEOMeGb/) between algebra-integral and blackhole is Quoter abi, which affect how we construct multicall callData and decodeFunction. I guess we can abstract this part with e.g. `getMultiCallData(from.address, to.address, pool.deployer, amount.toString())` and then in black-hole simply override this method for needed abi. 
It would prevent redundant duplication of whole `getPricingFromRpc` method

KanievskyiDanylo: image

KanievskyiDanylo (approved):

PR #1071, opened 2025-08-13
fix(BACK-2035): resolve PoolTracker data discrepancy 

Ivan011001:


mounibec:
@claude

src/dex/lite-psm/lite-psm.ts
```diff
@@ -490,51 +492,59 @@ export class LitePsm
...
+      const eventPool = this.eventPools[eventPoolKey];
+      const state = eventPool?.getStaleState();
+
+      if (!state) return null;
+
+      return {
+        exchange: this.dexKey,
+        address: config.psmAddress,
+        liquidityUSD: minLiq(state, config.gem.decimals),
+        connectorTokens: isDAI ? [config.gem] : [this.dai, this.usds],
```
KanievskyiDanylo: In previous implementation it was `isDaiOrUsds`, in current implementation it's only `isDAI`.Assume we should stick to `isDaiOrUsds`

PR #1078, opened 2025-09-04
fix(BACK-2052): transaction building issues

Ivan011001:


src/dex/uniswap-v2/ring-v2.ts
```diff
@@ -347,4 +348,44 @@ export class RingV2 extends UniswapV2 {
...
+    let functionName: RingV2Functions;
+
+    let ttl = 1200;
+    const deadline = `0x${(
+      Math.floor(new Date().getTime() / 1000) + ttl
+    ).toString(16)}`;
+
+    if (side === SwapSide.SELL) {
+      functionName = RingV2Functions.swapExactTokensForTokens;
+      args = [srcAmount, destAmount, data.path, data.router, deadline];
```
KanievskyiDanylo: Did you test it? I guess the 4th param (recipient) should be `augustusV5`, not the router address

Ivan011001: Updated recipient to `augustusV5` address

0xNazarii (approved):

PR #1081, opened 2025-09-11
cap adapter

prevostc:


src/dex/cap/cap.ts
```diff
@@ -0,0 +1,404 @@
...
+    recipient: Address,
+    data: CapData,
+    side: SwapSide,
+  ): DexExchangeParam {
+    const functionName = data.isMint ? 'mint' : 'burn';
+
+    return {
+      needWrapNative: this.needWrapNative,
+      dexFuncHasRecipient: true,
+      exchangeData: this.capIface.encodeFunctionData(functionName, [
```
KanievskyiDanylo, line 223: This method accepts only srcAmount, therefore it doesn't work well for BUY, so maybe it makes sense to support only SELL (I didn't find methods that accept destAmount which we can use for BUY)


[Example simulation](https://dashboard.tenderly.co/paraswap/paraswap/simulator/04335177-d84e-49ef-a9ad-d3c3375a0050/debugger?trace=0.4.0.1.1.0.1)
it takes all of the available slippage, which is not a good practice

src/dex/erc4626/config.ts
```diff
@@ -45,4 +45,10 @@ export const ERC4626Config: DexConfigMap<ERC4626Params> = {
       cooldownEnabled: true,
     },
   },
+  stcUSD: {
```
KanievskyiDanylo, line 48: please add tests for this new integration

PR #1242, opened 2026-09-15
perf(uniswap-v2): store only exchange and expiry in the pairs cache

0xNazarii:
## Why

The `dl_<network>_<dexKey>_pairs` Redis hashes written by the UniswapV2 base class total **8.12 GB across 39 keys — 79% of a 10.3 GB production ElastiCache instance**.

Sampling those hashes in production shows 98–100% of entries are *negative* records (the pair does not exist), averaging ~473 B of value payload. Almost all of that is two fully serialized `Token` objects that the reader never looks at: `_findPair` only ever reads `exchange` and `checkExistenceAfter`, and it already receives the real `token0`/`token1` from its caller. So the token data is pure redundancy.

## What

- Introduce `UniswapV2PairCacheRecord` (`{ exchange?, checkExistenceAfter }`) as the only thing persisted per pair, with `parsePairCacheRecord` / `isPairCacheRecordFresh` / `pairFromCacheRecord` helpers.
- `_findPair` rebuilds the returned `UniswapV2Pair` from its own token arguments plus the cached `exchange`. The in-memory pair shape, the `exchange || checkExistenceAfter > now` freshness rule, the `newlyCreatedPoolKeys` race guard, `this.pairs` population, and the in-flight `findPairPromises` dedup are all unchanged.
- Same treatment for the two other writers of these hashes: `Solidly._findSolidlyPairs` (which has its own `hmget`/`hmset` path on the same key, and rebuilds `stable` from values it already iterates) and `Camelot._findPair` (a verbatim copy of the logic with its own cache key).

Reads stay backward compatible: legacy fat entries are a strict superset of the new record, so they keep deserializing and the extra fields are ignored. A missing `checkExistenceAfter` defaults to `0`, which reads as already-expired — matching the previous `undefined &&` behaviour. Malformed JSON now degrades to a cache miss instead of throwing.

## Impact

Measured against `dl_8453_aerodrome_pairs` in production (966,324 fields / 684,042,476 bytes = 708 B per field):

| | today | after |
|---|---|---|
| value payload | ~473 B | 37 B |
| field name | ~95 B | ~95 B (unchanged) |
| Redis per-field overhead | ~140 B | ~140 B |
| **per entry** | **~708 B** | **~272 B** |

Roughly **60% off these hashes (~4.9 GB)** — but only as entries are rewritten. Negative records refresh within the 3-day `UNISWAP_V2_RECHECK_PAIR_EXISTENCE_AFTER_MS` window; positive entries are never rewritten (they short-circuit on `exchange`), so those keep their fat payload until deleted.

Note this does not address the underlying growth problem: unlike the UniswapV3 path, which sweeps its equivalent zset daily via `cleanExpiredNotExistingPoolsKeys`, the V2 `_pairs` hashes have no cleanup at all and entries are only removed when a pool is created. A separate sweep of expired negative entries is being run operationally, and a periodic sweeper for the V2 path is worth adding as a follow-up.

## Tests

New `src/dex/uniswap-v2/uniswap-v2-pairs-cache.test.ts` — 11 tests, no network. Covers: fresh negative record, expired negative (asserting the factory is called and exactly which fields get persisted), positive record with an expired timestamp, newly discovered pair, three legacy fat-shape cases (including that the caller's tokens win over stale cached copies), malformed JSON, and direct `parsePairCacheRecord` cases.

- `npx jest src/dex/uniswap-v2/uniswap-v2-pairs-cache.test.ts` → 11 passed
- `npx tsc --noEmit` → clean
- `eslint` + `prettier --check` on all changed files → clean

Not run: the existing uniswap-v2 / solidly / camelot integration and e2e suites, which require live RPC providers and forked-chain infra. Worth a CI run before merge given this touches the pair discovery path for every UniswapV2 fork.

---

> [!NOTE]
> **Medium Risk**
> Changes pair lookup and Redis serialization on a hot path for every Uniswap V2 fork; behavior is intended to match prior freshness rules with backward-compatible reads, but incorrect parsing could cause extra RPC or stale negatives until re-check.
> 
> **Overview**
> Shrinks Redis pair-discovery cache values for Uniswap V2–style DEXes by persisting only **`exchange`** and **`checkExistenceAfter`** instead of full serialized pairs (including redundant `Token` objects on negative “no pool” entries).
> 
> **Uniswap V2** adds `UniswapV2PairCacheRecord` plus `parsePairCacheRecord`, `isPairCacheRecordFresh`, and `pairFromCacheRecord`; `_findPair` rebuilds in-memory pairs from caller tokens and the minimal record, writes the slim shape on cache set, and treats bad JSON as a cache miss. **Solidly** and **Camelot** adopt the same read/write helpers on their shared `pairs` hash paths.
> 
> Legacy fat cache entries still deserialize (extra fields ignored). New unit tests in `uniswap-v2-pairs-cache.test.ts` cover freshness, re-check behavior, legacy shapes, and malformed entries. Version bumped to **5.1.17**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b834f3fe141eefb2879db3216d837d737d82b835. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


andriy-shymkiv (approved):

0xNazarii:
resolves DEXLIB-310

